PR #10, opened 2019-06-05
Add a pointer to our security vulnerability process.

geeknoid:


linsun:
I think we will need more info on how to fix them in a hidden fashion.  ultimately, the fix is recorded with issue and pr.

PR #14, opened 2019-07-18
Fix istio/lint bad substitution error

fpesce:
Despite working on my setup, and passing shellcheck, the SCRIPTPATH substitution did not work as intended in prow.
This fix has been tested on the prow pipeline, but I need to get it back to common-files source.

### DIFF
--- a/files/scripts/check_dockerfiles.sh
+++ b/files/scripts/check_dockerfiles.sh
@@ -23,13 +23,14 @@
 
 set -e
 
-SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SCRIPTPATH="$( cd "$( dirname "${0}")" && pwd -P)"
 ROOTDIR=$(dirname "${SCRIPTPATH}")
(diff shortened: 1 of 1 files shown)


geeknoid:
I'm worried that  this will now be different than what  we do in the other scripts in  the same directory. Should we change those too?

fpesce:
> I'm worried that this will now be different than what we do in the other scripts in the same directory. Should we change those too?

The BASH_SOURCE substitution seemed to fail (according to prow's log, that I was looking) so it might be a good idea.

@geeknoid: While working on that, I also noticed there seemed to be 2 different check_licence implementation in bin/linters.sh I was planning to fix this afterwards.

geeknoid:
Oh, I think the problem is that this script uses sh instead of bash. Could you switch to bash and see if we can then keep the BASH_SOURCES like the other scripts?

fpesce:
> Oh, I think the problem is that this script uses sh instead of bash. Could you switch to bash and see if we can then keep the BASH_SOURCES like the other scripts?

Indeed, good catch.

PR #15, opened 2019-07-25
Disable mergify, now using Tide.

geeknoid:


howardjohn:
doesnt look like tide is enabled yet

geeknoid:
@howardjohn 

There's definitely a  lag in enabling tide. But also, you're not in the CODEOWNERS file...

PR #16, opened 2019-07-30
Update go-bindata

geeknoid:
- Get the binary from the new container

- Pass user id to docker


geeknoid:
Force merge since Oz isn't around.

PR #20, opened 2019-08-06
Updates from operator into commmon-files

ericvn:
will be pulled back into operator in another PR

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -20,6 +20,7 @@
 # limitations under the License.
 
 RUN =
+GOBIN ?= ./out/bin
 
 # Set the environment variable BUILD_WITH_CONTAINER to use a container
 # to build the repo. The only dependencies in this mode are to have make and
--- a/files/Makefile
+++ b/files/Makefile
@@ -21,19 +21,44 @@
 
 RUN =
 
-# Set the enviornment variable BUILD_WITH_CONTAINER to use a container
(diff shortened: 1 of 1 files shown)


files/Makefile
```diff
@@ -20,20 +20,46 @@
...
 # docker. If you'd rather build with a local tool chain instead, you'll need to
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 ifeq ($(BUILD_WITH_CONTAINER),1)
-IMG = gcr.io/istio-testing/build-tools:2019-08-05
+IMG = gcr.io/istio-testing/build-tools:2019-08-06
 UID = $(shell id -u)
 PWD = $(shell pwd)
 GOBIN_SOURCE ?= $(GOPATH)/bin
 GOBIN ?= /work/out/bin
```
sdake: this line can be replaced by line 23.  I'm still not sure this is correct long term.  I think we need to bindmount ${GOPATH}/bin in `rshared` mode so we can write out binaries.  But we can do this as a next step.

sdake:
I approved this PR - not as a codeowner but tide merged it anyway.  It looks like our infrastructure (was) setup to permit admins to approve any PR which was unintended.  @geeknoid has got this fixed from my understanding.  See screenshot for evidence of tide misbehavior.

Screen Shot 2019-08-06 at 11 38 54 PM

PR #25, opened 2019-08-09
Fix the generated .commonfiles.sha

ericvn:
The generated .commonfiles.sha for a `make updatecommon` ends up containing the commit id of the repo where the update was happening. I assume it should contain the commit id of the common-files repo that was actually used.


ericvn:
It should be noted that one needs to run `make update common` twice to get things inline since the existing _Makefile.common.mk_ is still the old one and will generate the old value.

PR #35, opened 2019-08-20
Bunch of updates

geeknoid:
- Add the -i option to docker run so that CTRL-C behaves correctly when
running the build tools container.

- Remove dates for our copyright notices

- Add common targets for functionality delivered by the build container.
These targets will be leveraged by the repo-specific makefiles to crete
their own high-level targets. For exmple, a rpeo-specific `lint` target
might call depend on `lint-go` and `lint-copyright-banner`.


files/Makefile.common.mk
```diff
@@ -19,9 +19,32 @@
...
 
-updatecommon:
+lint-dockerfiles:
+	@find . -path ./vendor -prune -o -type f -name 'Dockerfile*' -print0 | xargs -0 hadolint -c ./.hadolint.yml
+
+lint-scripts:
+	@find . -path ./vendor -prune -o -type f -name '*.sh' -print0 | xargs -0 shellcheck
+
+lint-yaml:
+	@find . -path ./vendor -prune -o -type f -name '*.yml' -print0 | xargs -0 yamllint -c ./.yamllint.yml
```
howardjohn: do we want all these config files to be top level in the repos?

PR #51, opened 2019-09-01
Add support to lint the common files.

geeknoid:
- Also, stop linting generated Python files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+lint:
+	@cp files/common/Makefile.common.mk files/Makefile.core.mk
+	@echo >files/Makefile.overrides.mk "BUILD_WITH_CONTAINER ?= 1"
```
howardjohn: why not just BUILD-WITH-CONTAINER=1 make -f files/common/Makefile.core? are the paths messed up if you do that?

geeknoid: Yes, the paths are messed up. I tried a  variety of combinations, but I couldn't get everything happy.

PR #58, opened 2019-09-06
Use realpath instead of readlink

jwendell:
Output of readlink on linux is something like
```
$ readlink /etc/localtime
../usr/share/zoneinfo/America/New_York
```

Which breaks the `-v` docker cli option

realpath works fine:
```
$ realpath /etc/localtime
/usr/share/zoneinfo/America/New_York
```


rcernich:
I don't think realpath is available on macos.  #66 incorporates an appropriate fix for this, which should work on both macos and rhel/fedora.

sdake:
Thank you for the submission.  This is not portable to macOS.  We have a strong vocal contingent of developers that use macOS exclusively so this or is not the right approach.

Thanks again.

If you are satisfied with this review please close and comment.

jwendell:
ok, fair enough. I don't have macos, and I asked two people who have it and they confirmed realpath was installed on their systems. readlink -f (from the other pr) works for me too.

PR #64, opened 2019-09-11
Ensure GOBIN is set to an absolute path as this is required in istio/…

geeknoid:
…istio.


geeknoid:
This specific change shouldn't impact the path relative to what it was I think. I'm just replacing . with $(pwd) which I think translates to the same actual location, right?

PR #66, opened 2019-09-11
support use of alternate container tool

rcernich:
fix issue with localtime mount

allow users to specify IMG for local development

Signed-off-by: rcernich <rcernich@redhat.com>

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -29,7 +29,9 @@
 export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-IMG = gcr.io/istio-testing/build-tools:2019-09-17T18-24-15
+CONTAINER_CLI ?= docker
+DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
+IMG ?= gcr.io/istio-testing/build-tools:2019-09-17T18-24-15
 UID = $(shell id -u)
 PWD = $(shell pwd)
 GOBIN_SOURCE ?= $(GOPATH)/bin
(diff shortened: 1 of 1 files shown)


files/Makefile
```diff
@@ -29,7 +29,9 @@
 export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-IMG = gcr.io/istio-testing/build-tools:2019-09-11T09-52-48
+CONTAINER_CLI ?= docker
+CONTAINER_OPTIONS ?= -v /var/run/docker.sock:/var/run/docker.sock
```
jwendell: It seems this variable was already being used before this PR for other purpose. Thus this line would break current behavior. What about calling this variable `CONTAINER_EXTRA_OPTIONS`?

files/Makefile
```diff
@@ -59,14 +61,13 @@ endif
...
-RUN = docker run -t -i --sig-proxy=true -u $(UID) --rm \
+RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID) --rm \
 	-e GOOS="$(GOOS)" \
 	-e GOARCH="$(GOARCH)" \
 	-e GOBIN="$(GOBIN)" \
 	-e BUILD_WITH_CONTAINER="$(BUILD_WITH_CONTAINER)" \
 	-v /etc/passwd:/etc/passwd:ro \
-	-v $(readlink /etc/localtime):/etc/localtime:ro \
-	-v /var/run/docker.sock:/var/run/docker.sock \
+	-v $(shell readlink -f /etc/localtime):/etc/localtime:ro \
```
jwendell: is `readlink -f` available on mac?

sdake: no:
```
Stevens-MacBook-Pro:~ sdake$ readlink -f
readlink: illegal option -- f
```

There are two options here that are viable.  We can either write our own readlink in bash or detect the operating system environment we are running in and set our readlink operation in that manner.  We already detect our operating system on line 57, so that may be the least painless way.  I'd use something like "OS_MOUNT" and set to `-v $(shell readlink -f /etc/localtime):/etc/localtime:ro)` or alternatively set to `-v $(shell readliink /etc/localtime):/etc/localtime:ro)`.

rcernich: okay, preserving `readlink` for darwin and using `readlink -f` for linux.

rcernich:
This should be good to go now.

rcernich:
@geeknoid, mind giving this a quick review, so it can get merged?  TIA

PR #67, opened 2019-09-12
Update golangci-lint to v1.18.0

howardjohn:
This matches what is in istio/istio

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -29,7 +29,7 @@
 export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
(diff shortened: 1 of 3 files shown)


geeknoid:
as part of this change, could you also update run_lint.sh to include GOGC=25 in the command-line so that we can successfully lint istio/istio?

Thanks.

howardjohn:
@geeknoid made the updates. I noticed the istio/tools docker build is failing though https://prow.istio.io/job-history/istio-prow/logs/tools-containers

geeknoid:
@howardjohn Yeah, I never did get that job working... I'll look into it. The container is building, it's just failing when trying to push it go gcr.io:

denied: Access denied.

PR #69, opened 2019-09-19
Set default make goal

SataQiu:
I tried to build the code accroding to https://github.com/istio/istio/wiki/Using-the-Code-Base#building-the-code, but there was an error:
```
make[1]: Nothing to be done for 'bin'.
```
I think this is because there is no default make goal.
This PR will fix this issue.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -19,6 +19,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+.DEFAULT_GOAL := default
+
 # allow optional per-repo overrides
 -include Makefile.overrides.mk
 


sdake:
@SataQiu can you elaborate on the repository you were attempting to build?

Cheers
-steve

SataQiu:
@sdake  It is `istio/istio`.

SataQiu:
Hi @sdake 
I think the problem may be caused by this code: https://github.com/istio/istio/blob/master/Makefile.overrides.mk#L19
```
# create phony targets for the top-level items in the repo
PHONYS := $(shell ls | grep -v Makefile)
.PHONY: $(PHONYS)
$(PHONYS):
	@$(MAKE) $@
```
This will make the first goal as `make bin`.
And that's not an valid goal.

Can you help me?

sdake:
@SataQiu I've got a bunch of outstanding requests to fix brokeness in the build system.  I am working on it as fast as possible :) . I don't have an immediate answer to your question however.

Cheers
-steve

SataQiu:
> 
> 
> I think the problem may be caused by this code: https://github.com/istio/istio/blob/master/Makefile.overrides.mk#L19
> 
> ```
> # create phony targets for the top-level items in the repo
> PHONYS := $(shell ls | grep -v Makefile)
> .PHONY: $(PHONYS)
> $(PHONYS):
> 	@$(MAKE) $@
> ```
> 
> This will make the first goal as `make bin`.
> And that's not a valid goal.
> 

@geeknoid WDYT?

geeknoid:
I think the DEFAULT_GOAL value should probably be set in the Makefile.overrides.mk file in istio/istio. Then we can set different values for different repos.

SataQiu:
close this in favor of  istio/istio#17370

PR #71, opened 2019-09-20
Update container image.

geeknoid:
- Switch to latest build-tools container.

- Integrate changes from istio/istio/bin/gobuild.sh into the common gobuild.sh so that I can
delete the dedicated istio/istio version.

- Fix problems with readlink usage.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -31,7 +31,7 @@ export BUILD_WITH_CONTAINER ?= 0
 ifeq ($(BUILD_WITH_CONTAINER),1)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
-IMG ?= gcr.io/istio-testing/build-tools:2019-09-17T18-24-15
+IMG ?= gcr.io/istio-testing/build-tools:2019-09-20T15-04-58
 UID = $(shell id -u)
 PWD = $(shell pwd)
 GOBIN_SOURCE ?= $(GOPATH)/bin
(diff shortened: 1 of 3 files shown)


files/common/scripts/gobuild.sh
```diff
@@ -26,16 +33,17 @@ fi
 SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 OUT=${1:?"output path"}
-shift
+BUILDPATH=${2:?"path to build"}
```
howardjohn: this is undoing the changes on istio/istio. I already merged them in. The LDFLAGS one is new though

files/common/scripts/gobuild.sh
```diff
@@ -65,4 +74,5 @@ time GOOS=${BUILD_GOOS} GOARCH=${BUILD_GOARCH} ${GOBINARY} build \
         ${V} "${GOBUILDFLAGS_ARRAY[@]}" ${GCFLAGS:+-gcflags "${GCFLAGS}"} \
         -o "${OUT}" \
         -trimpath \
-        -ldflags "${LDFLAGS} ${LD_EXTRAFLAGS}" "${@}"
```
howardjohn: this part is reverting my change as well -- should be `@` not `BUILDPATH`

files/Makefile
```diff
@@ -69,7 +69,7 @@ RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID) --rm \
 	-e GOBIN="$(GOBIN)" \
 	-e BUILD_WITH_CONTAINER="$(BUILD_WITH_CONTAINER)" \
 	-v /etc/passwd:/etc/passwd:ro \
-	-v $(shell $(read_link) /etc/localtime):/etc/localtime:ro \
+	-v $(read_link):/etc/localtime:ro \
```
sdake: I'm not convinced this is working as you are expecting.  can you run echo $(read_link) and see what is printed?

geeknoid: This produces the right command-line. Docker runs happily.

files/common/scripts/gobuild.sh
geeknoid: Sorry, I guess I must've gotten confused with the different versions. Please see the latest update for a more sane view I hope.

files/Makefile
geeknoid: Shoot, you're right. Argh.

PR #72, opened 2019-09-22
Set timezone data in a platform independent way

sdake:
Mounting /etc/timezone is not correct across platforms and can lead
to corruption of the timezone database.  This is why Docker does not
permit the timezone file to be mounted into the container.  Instead,
pass an Environment variable containing `TZ`. Most modern apps
understand how to process `TZ`.  For those that don't, I plan to
introduce an entrypoint to the docker command to overwrite /etc/localtime
with a new file that contains the contents of this environment variable.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -53,13 +53,19 @@ GOARCH ?= $(GOARCH_LOCAL)
 LOCAL_OS := $(shell uname)
 ifeq ($(LOCAL_OS),Linux)
    GOOS_LOCAL = linux
-   read_link=$(readlink -f /etc/localtime)
+   # Apparently Linux needs a -f flag.
+   READLINK_FLAGS="-f"
 else ifeq ($(LOCAL_OS),Darwin)
    GOOS_LOCAL = darwin
-   read_link=$(readlink /etc/localtime)
+   READLINK_FLAGS=""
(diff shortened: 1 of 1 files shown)


sdake:
inspiration taken from this Docker issue: https://github.com/moby/moby/issues/12084

sdake:
This environment variable makes it all the way into Makefile.core.mk \o/.

```
Stevens-MacBook-Pro:istio sdake$ !make
make build
echo TZ=`readlink "" /etc/localtime | sed -e 's/^.*zoneinfo\///'`
TZ=America/Phoenix
warning: GOPATH set to GOROOT (/go) has no effect
warning: GOPATH set to GOROOT (/go) has no effect
echo "TIMEZONE=America/Phoenix"
TIMEZONE=America/Phoenix
touch: cannot touch '/go/bin/have_go_1.10': Permission denied
Makefile.core.mk:271: recipe for target '/go/bin/have_go_1.10' failed
make: *** [/go/bin/have_go_1.10] Error 1
make: *** [build] Error 2
```

The first printing of TZ is from the `Makefile` itself, the second is from `Makefile.core.mk` from istio/istio with `BUILD_WITH_CONTAINER=1`.

PR #77, opened 2019-09-23
Adapt to latest build tools container.

geeknoid:
- No longer pass GOARCH/GOOS to docker, we don't need that when
building from the container.


geeknoid:
Indeed. I think what we need then is to have the user define an environment variable where build outputs are deposited. We just don't have a good place for that at the moment. We need to treat "artifacts output for the sake of the build itself" from "final artifacts for shipping to customers". /home/out is for the first case, we need to also support the second.

PR #78, opened 2019-09-24
verify go version before build

SataQiu:
Ref: istio/istio#17298

### DIFF
--- a/files/common/scripts/gobuild.sh
+++ b/files/common/scripts/gobuild.sh
@@ -75,7 +75,7 @@ done < "${BUILDINFO}"
 minimum_go_version=go1.13 # supported patterns: go1.x, go1.x.x (x should be a number)
 IFS=" " read -ra go_version <<< "$(${GOBINARY} version)"
 if [[ "${minimum_go_version}" != $(echo -e "${minimum_go_version}\n${go_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) && "${go_version[2]}" != "devel" ]]; then
-    echo "Warning: Detected that you are using a low version of GO. Istio requires ${minimum_go_version} or greater."
+    echo "Warning: Detected that you are using an older version of the Go compiler. Istio requires ${minimum_go_version} or greater."
 fi
 
 time GOOS=${BUILD_GOOS} GOARCH=${BUILD_GOARCH} ${GOBINARY} build \
--- a/files/common/scripts/gobuild.sh
+++ b/files/common/scripts/gobuild.sh
@@ -70,6 +70,14 @@ while read -r line; do
     LD_EXTRAFLAGS="${LD_EXTRAFLAGS} -X ${line}"
 done < "${BUILDINFO}"
 
+# verify go version before build
+# NB. this was copied verbatim from Kubernetes hack
+minimum_go_version=go1.13 # supported patterns: go1.x, go1.x.x (x should be a number)
+IFS=" " read -ra go_version <<< "$(${GOBINARY} version)"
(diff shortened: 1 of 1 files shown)


files/common/scripts/gobuild.sh
```diff
@@ -70,6 +70,14 @@ while read -r line; do
     LD_EXTRAFLAGS="${LD_EXTRAFLAGS} -X ${line}"
 done < "${BUILDINFO}"
 
+# verify go version before build
+# NB. this was copied verbatim from Kubernetes hack
+minimum_go_version=go1.13 # supported patterns: go1.x, go1.x.x (x should be a number)
+IFS=" " read -ra go_version <<< "$(${GOBINARY} version)"
+if [[ "${minimum_go_version}" != $(echo -e "${minimum_go_version}\n${go_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) && "${go_version[2]}" != "devel" ]]; then
+    echo "Warning: Detected that you are using a low version of GO. Istio requires ${minimum_go_version} or greater."
```
geeknoid: Is outputting a warning sufficient/useful? Seems like we should just stop the build instead.

geeknoid:
```suggestion
    echo "Warning: Detected that you are using an older version of the Go compiler. Istio requires ${minimum_go_version} or greater."
```

howardjohn: the buiild is already going to fail if they don't use 1.13. I don't think there is any reason to fail early, since this check may have false positives

files/common/scripts/gobuild.sh
```diff
@@ -70,6 +70,14 @@ while read -r line; do
     LD_EXTRAFLAGS="${LD_EXTRAFLAGS} -X ${line}"
 done < "${BUILDINFO}"
 
+# verify go version before build
+# NB. this was copied verbatim from Kubernetes hack
+minimum_go_version=go1.13 # supported patterns: go1.x, go1.x.x (x should be a number)
+IFS=" " read -ra go_version <<< "$(${GOBINARY} version)"
```
ericvn: -a is not a valid option for `read` on MacOS. Not sure if we expect users to be able to build locally or not.

ericvn: OK. Did some testing of only this read with/without the `-a` on the Mac (and maybe I need to also add using zsh)

with -a and adding an `echo go_version ${go_version}`
```
go_version go
```

Removing the -a:
```
go_version go version go1.13 darwin/amd64
Warning: Detected that you are using a low version of GO. Istio requires go1.13 or greater.
```

SataQiu: Thanks for your feed back. @ericvn 
I'm sorry I don't have a Mac.
Can you help me test the following command again?
```sh
# IFS=" " read -ra go_version <<< "$(go version)"
# echo ${go_version[@]} 
# echo ${#go_version[@]}
```

The `echo ${go_version}`  command outputs only the first element of the array.
We should use `echo ${go_version[@]}` to output all the elements.

ericvn: Output:
```
./test.sh
go version go1.13 darwin/amd64
4
```
So looks like on a MacOS (10.14.6 anyway), the `-a` while yield the correct item.

More experimentation on my Mac determines that the reason I see the warning is the `-e` on the echo. If I output the generated string being compared surrounded by x's I have:
```
./test.sh
go version go1.13 darwin/amd64
4
x-e go1.13x
``` 

Removing the -e from the echo prevents the warning on my Mac. I suspect that removing the -e while cause issues on Linux :(.

PR #79, opened 2019-09-25
Report the type of build environment in use.

sdake:


files/Makefile
```diff
@@ -55,6 +55,7 @@ REPO_NAME = $(shell basename $(REPO_ROOT))
 TARGET_OUT ?= $(HOME)/istio_out/$(REPO_NAME)
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
+$(info Building with Istio's cononical build container.)
```
geeknoid:
```suggestion
$(info Building with Istio's canonical build container.)
```

geeknoid: How about include the date of the container in this text?

sdake: I thought of that - though it may be too long - will add and see how it looks on nthe terminal - thanks :)

sdake: PTAL.

PR #91, opened 2019-10-02
Pass BUILD_WITH_CONTAINER to the child containers

sdake:
The per-repo makefiles environment in some cases needs to know if they
are being built in the build-container.  To provide that information, pass
it in via the environment.

Depends-On: https://github.com/istio/tools/pull/402

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -57,7 +57,7 @@ TARGET_OUT ?= $(HOME)/istio_out/$(REPO_NAME)
 ifeq ($(BUILD_WITH_CONTAINER),1)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
-IMG ?= gcr.io/istio-testing/build-tools:2019-10-01T21-32-16
+IMG ?= gcr.io/istio-testing/build-tools:2019-10-02T14-57-08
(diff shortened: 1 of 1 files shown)


geeknoid:
I took this setting out last week explicitly since it was leading to unexpected DIND. Something was recursively calling 'make' on the top-level makefile from within build-tools, and ended up calling docker again, which is definitely not the intent.

If we can ensure this recursion will never happen by having nested cases always invoke `make -f Makefile.core.mk` instead, then we can reintroduce this I suppose.

An alternative is to use a different name. Something like IN_BUILD_CONTAINER for example.

sdake:
> I took this setting out last week explicitly since it was leading to unexpected DIND. Something was recursively calling 'make' on the top-level makefile from within build-tools, and ended up calling docker again, which is definitely not the intent.
> 
> If we can ensure this recursion will never happen by having nested cases always invoke `make -f Makefile.core.mk` instead, then we can reintroduce this I suppose.
> 
> An alternative is to use a different name. Something like IN_BUILD_CONTAINER for example.

which repository?  I am not opposed to using a different environment variable although it does expand the footprint of the makefile api.

sdake:
PTAL - used variable as you suggested.

geeknoid:
I think this was happening in istio/istio...

PR #94, opened 2019-10-02
extract BUILD_WITH_CONTAINER docker command into a script

XanderStrike:
Also adds make shell to drop the user to a bash prompt within the
container.

resolves istio/istio#17476 - all commands arguments to run-docker.sh are passed
into the container

resolves istio/istio#17474 - the script will pass along a filtered picture of the host's environment to the container

Notable changes and omissions:

* When the script is run in isolation, it uses the `latest` tag of
build-tools, the makefile overrides it with the dated tag. Hopefully
someone's in charge of keeping that up to date.
~~* I left off --sig-proxy=true which helps istio/istio#17476, but does not resolve
it fully. I believe it was causing problems because the entrypoint is
PID 1 which has some special rules for SIGINTs.~~ _I was incorrect here, it's added back in_

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -28,60 +28,11 @@
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
-LOCAL_ARCH := $(shell uname -m)
-ifeq ($(LOCAL_ARCH),x86_64)
-    TARGET_ARCH ?= amd64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
-    TARGET_ARCH ?= arm64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 4),armv)
-    TARGET_ARCH ?= arm
-else
-    $(error This system's architecture $(LOCAL_ARCH) isn't supported)
-endif
-
-LOCAL_OS := $(shell uname)
-ifeq ($(LOCAL_OS),Linux)
-    TARGET_OS ?= linux
-    READLINK_FLAGS="-f"
-else ifeq ($(LOCAL_OS),Darwin)
-    TARGET_OS ?= darwin
-    READLINK_FLAGS=""
-else
-    $(error This system's OS $(LOCAL_OS) isn't supported)
-endif
-
-export TARGET_OUT ?= $(shell pwd)/out/$(TARGET_ARCH)_$(TARGET_OS)
-
 ifeq ($(BUILD_WITH_CONTAINER),1)
-export TARGET_OUT = /work/out/$(TARGET_ARCH)_$(TARGET_OS)
-CONTAINER_CLI ?= docker
-DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
 IMG ?= gcr.io/istio-testing/build-tools:2019-10-11T13-37-52
-UID = $(shell id -u)
-PWD = $(shell pwd)
-
-$(info Building with the build container: $(IMG).)
-
-# Determine the timezone across various platforms to pass into the
-# docker run operation. This operation assumes zoneinfo is within
-# the path of the file.
-TIMEZONE=`readlink $(READLINK_FLAGS) /etc/localtime | sed -e 's/^.*zoneinfo\///'`
-
-RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):docker --rm \
-	-e IN_BUILD_CONTAINER="$(BUILD_WITH_CONTAINER)" \
-	-e TZ="$(TIMEZONE)" \
-	-e TARGET_ARCH="$(TARGET_ARCH)" \
-	-e TARGET_OS="$(TARGET_OS)" \
-	-e TARGET_OUT="$(TARGET_OUT)" \
-	-e HUB="$(HUB)" \
-	-e TAG="$(TAG)" \
-	-v /etc/passwd:/etc/passwd:ro \
-	$(DOCKER_SOCKET_MOUNT) \
-	$(CONTAINER_OPTIONS) \
-	--mount type=bind,source="$(PWD)",destination="/work" \
-	--mount type=volume,source=go,destination="/go" \
-	--mount type=volume,source=gocache,destination="/gocache" \
-	-w /work $(IMG)
+export IMG
+$(info Building with the build container: $(IMG))
+RUN = ./common/scripts/run-docker.sh
 else
 $(info Building with your local toolchain.)
 RUN =
@@ -90,10 +41,25 @@ endif
 
 MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
 
+# If these variables are defined in the overrides, exporting them adds them to
+# the environment of the sub-make
+export TARGET_ARCH
+export TARGET_OS
+export TARGET_OUT
+export CONTAINER_CLI
+export DOCKER_SOCKET_MOUNT
+export CONTAINER_OPTIONS
+export TZ
+export HUB
+export TAG
+
 %:
 	@$(MAKE) $@
 
 default:
 	@$(MAKE)
+
+shell:
+	@$(RUN) /bin/bash
 
 .PHONY: default
--- a/files/Makefile
+++ b/files/Makefile
@@ -29,8 +29,6 @@
 export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-IMG ?= gcr.io/istio-testing/build-tools:2019-10-11T13-37-52
-export IMG
 $(info Building with the build container: $(IMG))
 RUN = ./common/scripts/run-docker.sh
 else
--- a/files/common/scripts/run-docker.sh
+++ b/files/common/scripts/run-docker.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# WARNING: DO NOT EDIT, THIS FILE IS PROBABLY A COPY
+#
+# The original version of this file is located in the https://github.com/istio/common-files repo.
+# If you're looking at this file in a different repo and want to make a change, please go to the
+# common-files repo, make the change there and check it in. Then come back to this repo and run
+# "make update-common".
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# set -ex
+
+local_arch=$(uname -m)
+if [[ $local_arch == x86_64 ]]; then
+    target_arch=amd64
+elif [[ $local_arch == armv8* ]]; then
+    target_arch=arm64
+elif [[ $local_arch == armv* ]]; then
+    target_arch=arm
+else
+    echo "This system's architecture, $local_arch, isn't supported"
+    exit 1
+fi
+
+local_os=$(uname)
+if [[ $local_os == Linux ]]; then
+    target_os=linux
+    readlink_flags="-f"
+elif [[ $local_os == Darwin ]]; then
+    target_os=darwin
+    readlink_flags=""
+else
+    echo "This system's OS, $local_os, isn't supported"
+    exit 1
+fi
+
+timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
+
+out="${TARGET_OUT:/work/out/${target_arch}_${target_os}}"
+
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+
+container_cli="${CONTAINER_CLI:-docker}"
+
+$container_cli pull "$image"
+
+# $CONTAINER_OPTIONS becomes an empty arg when quoted, so SC2086 is disabled for the
+# following command only
+# shellcheck disable=SC2086
+$container_cli run -it --rm -u "$(id -u):docker" \
+    --sig-proxy=true \
+    ${DOCKER_SOCKET_MOUNT:+"-v /var/run/docker.sock:/var/run/docker.sock"} \
+    -v /etc/passwd:/etc/passwd:ro \
+    $CONTAINER_OPTIONS \
+    -e IN_BUILD_CONTAINER=1 \
+    -e TZ="${timezone:-$TZ}" \
+    -e TARGET_ARCH="${TARGET_ARCH:-$target_arch}" \
+    -e TARGET_OS="${TARGET_OS:-$target_os}" \
+    -e TARGET_OUT="$out" \
+    -e HUB="$HUB" \
+    -e TAG="$TAG" \
+    --mount "type=bind,source=$(pwd),destination=/work" \
+    --mount "type=volume,source=go,destination=/go" \
+    --mount "type=volume,source=gocache,destination=/gocache" \
+    -w /work "$image" "$@"
+
--- a/files/common/scripts/run-docker.sh
+++ b/files/common/scripts/run-docker.sh
@@ -49,29 +49,30 @@ fi
 
 timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
 
-out="${TARGET_OUT:/work/out/${target_arch}_${target_os}}"
+out="${TARGET_OUT:-/work/out/${target_arch}_${target_os}}"
 
-image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+image="${IMG:-gcr.io/istio-testing/build-tools:2019-10-11T13-37-52}"
 
 container_cli="${CONTAINER_CLI:-docker}"
 
-$container_cli pull "$image"
+parent_env="-e \"$(env | grep -v "^_\|PATH\|SHELL\|EDITOR\|TMUX\|USER\|HOME\|PWD\|TERM\|GO\|rvm\|SSH" | awk 'ORS="\" -e \""' | sed 's/ -e "$//')"
+
+"$container_cli" pull "$image"
 
 # $CONTAINER_OPTIONS becomes an empty arg when quoted, so SC2086 is disabled for the
 # following command only
 # shellcheck disable=SC2086
-$container_cli run -it --rm -u "$(id -u):docker" \
+"$container_cli" run -it --rm -u "$(id -u):docker" \
     --sig-proxy=true \
     ${DOCKER_SOCKET_MOUNT:+"-v /var/run/docker.sock:/var/run/docker.sock"} \
     -v /etc/passwd:/etc/passwd:ro \
     $CONTAINER_OPTIONS \
+    "$parent_env" \
     -e IN_BUILD_CONTAINER=1 \
     -e TZ="${timezone:-$TZ}" \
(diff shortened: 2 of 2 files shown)


files/Makefile
```diff
@@ -28,58 +28,12 @@
...
-else ifeq ($(LOCAL_OS),Darwin)
-   TARGET_OS ?= darwin
-   READLINK_FLAGS=""
-else
-   $(error "This system's OS $(LOCAL_OS) isn't recognized/supported")
-endif
-
-REPO_ROOT = $(shell git rev-parse --show-toplevel)
-REPO_NAME = $(shell basename $(REPO_ROOT))
 TARGET_OUT ?= $(HOME)/istio_out/$(REPO_NAME)
```
geeknoid: REPO_NAME isn't defined, so this is not setting TARGET_OUT to the desired value.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+repo_name=$(basename "$repo_root")
+target_out="${TARGET_OUT:-$HOME/istio_out/$repo_name}"
+
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+
+docker_sock="${DOCKER_SOCKET_MOUNT:-"-v /var/run/docker.sock:/var/run/docker.sock"}"
+
+docker pull "$image"
+
+docker run -it --rm -u "$(id -u):docker" \
```
geeknoid: The reason we used --sig-proxy=true is so that issuing a ctrl-c in the host will cause computations in the container to halt. Otherwise, I think there's no easy way for a developer to stop a long-running activity such as a test run. Seems like we need to keep this option.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+    -u root \
+    --cap-add=NET_ADMIN \
```
geeknoid: @sdake Will this work with all the work you've been doing?

sdake: this may work - although it isn't correct.  -u is specified in line 62.  No need to specify it again as `-u root`.  the `--cap-app=NET_ADMIN` hasn't been necessary for me even without running as root.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+
+timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
+
+repo_root=$(git rev-parse --show-toplevel)
+repo_name=$(basename "$repo_root")
+target_out="${TARGET_OUT:-$HOME/istio_out/$repo_name}"
+
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+
+docker_sock="${DOCKER_SOCKET_MOUNT:-"-v /var/run/docker.sock:/var/run/docker.sock"}"
```
sdake: I believe in the original Makefile this is set based upon the value of CONTAINER_TOOLS.  This permits non-docker runtimes to build Istio (such as Podman).

files/Makefile
```diff
@@ -28,58 +28,12 @@
...
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-CONTAINER_CLI ?= docker
```
sdake: this logic was lost.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+    readlink_flags="-f"
+elif [[ $local_os == Darwin ]]; then
+    target_os=darwin
+    readlink_flags=""
+else
+    echo "This system's OS, $local_os, isn't recoginized/supported"
+    exit 1
+fi
+
+timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
```
sdake: this doesn't work correctly - atlaest in the makefile.  Perrhaps the piping works in bash.  can you verify by printing the value of timezone with echo $timezone?

Also variables should be capitalized in a script imo.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+docker pull "$image"
+
+docker run -it --rm -u "$(id -u):docker" \
+    -u root \
+    --cap-add=NET_ADMIN \
+    "$docker_sock" \
+    -v /etc/passwd:/etc/passwd:ro \
+    "$CONTAINER_OPTIONS" \
+    -e IN_BUILD_CONTAINER=1 \
+    -e WHAT="$WHAT" \
```
sdake: why this variable WHAT?

sdake: @howardjohn was the idea to pass `WHAT` in that issue you opened, or to pass all environment variables?  We could pass all environment variables (by using env to determine the environment, then filtering those that are essential to the container runtime out of the env list, then producing a list of `-e e envvar=$"$envvar".  Can you help set requirements?

Cheers
-steve

files/common/scripts/run-docker.sh
sdake: @geeknoid I was wondering why that was there. :) - Cheers :)

files/common/scripts/run-docker.sh
howardjohn: I don't really care about WHAT actually - I would never use it if we have the script since then I could just do `./docker-script go test ./foo`. But there are tons of other things that require env variables - HUB, TAG, REFRESH_GOLDEN, etc. I don't think we should whitelist, it may be better to instead blacklist env variables that will break the container -- I am not sure what exactly those are.

If its not feasible to do this maybe we can have some CONTAINER_ENV var and we pass everything in that, although that makes the dev UX worse

sdake:
> Also adds make shell to drop the user to a bash prompt within the
> container.
> 
> resolves [istio/istio#17476](https://github.com/istio/istio/issues/17476) - all commands arguments to run-docker.sh are passed
> into the container
> 
> resolves [istio/istio#17474](https://github.com/istio/istio/issues/17474) - only for the case of WHAT though, is there documentation on
> other environment variables that can be passed to make?
> 
> Notable changes and omissions:
> 
> * When the script is run in isolation, it uses the `latest` tag of
>   build-tools, the makefile overrides it with the dated tag. Hopefully
>   someone's in charge of keeping that up to date.

Istio's CI system automatically builds and tags in https://gcr.io/istio-testing on every merge of a successfully building dockerfile.

> * I left off --sig-proxy=true which helps [istio/istio#17476](https://github.com/istio/istio/issues/17476), but does not resolve
>   it fully. I believe it was causing problems because the entrypoint is
>   PID 1 which has some special rules for SIGINTs.

It sounds like you mean to say that the Makefile operation's signals are not being processed by `docker-run.sh`.  Is that accurate.  If not could you expand?

files/common/scripts/run-docker.sh
XanderStrike: Interesting, I thought I observed the opposite behavior during `make docker` but I think it's only during parts of that target. I'll add it back in.

files/Makefile
XanderStrike: Good catch! Thank you for this.

files/common/scripts/run-docker.sh
XanderStrike: The one-liner works on my mac development machine and on my Ubuntu 18.04 build machine.

BASH convention is environment variables and internal shell variables are capitalized, and variables local to scripts or functions are lowercase. This way script authors do not have to handle the case when the variable names they want to use are already assigned. Since these variables are local to the script and not exported, I believe they should stay lowercase.

files/Makefile
XanderStrike: Good catch, restored.

files/common/scripts/run-docker.sh
XanderStrike: [It depended on an environment variable](https://github.com/istio/common-files/blob/master/files/Makefile#L59) which is consistent with what's done here. Podman, for example, has no need for a socket since there's no daemon AFAIK.

files/common/scripts/run-docker.sh
XanderStrike: FWIW I think it'd be a lot easier to come up with a list of env variables that actually do something. I'm surprised/disappointed such a list doesn't already exist somewhere, or at the very least that I was unable to find it.

I think the CONTAINER_OPTIONS var could serve the purpose of CONTAINER_ENV, since it's a bad experience either way it doesn't cost any extra to tell people to put `-e ` at the beginning.

files/common/scripts/run-docker.sh
XanderStrike: Good call, that was a mistake merging my script with the makefile command. Removed.

XanderStrike:
Thanks for that context and the review @sdake, I've added `--sig-proxy=true` back in because I think my assumptions were incorrect. I'm glad to hear IMG is updated automatically, hopefully whatever updates the makefile won't need to change.

files/common/scripts/run-docker.sh
sdake: > I don't really care about WHAT actually - I would never use it if we have the script since then I could just do `./docker-script go test ./foo`. But there are tons of other things that require env variables - HUB, TAG, REFRESH_GOLDEN, etc. I don't think we should whitelist, it may be better to instead blacklist env variables that will break the container -- I am not sure what exactly those are.
> 
> If its not feasible to do this maybe we can have some CONTAINER_ENV var and we pass everything in that, although that makes the dev UX worse

right, I think we said the same thing differently. ..

sdake: I think we should run `env` and create a list of environment variables, and remove any needed for the build system to function properly.  There are not many of these (anything that is a current -e should be filtered out).

sdake:
> Thanks for that context and the review @sdake, I've added `--sig-proxy=true` back in because I think my assumptions were incorrect. I'm glad to hear IMG is updated automatically, hopefully whatever updates the makefile won't need to change.

humans update the makefile first by committing it to istio.io/common-files/common/Makefile - then per repo running `make update-common`.  We are ok with out of date images in the 15 or so repos we have at the moment as we finish out the build system.  Eventually, we will reach a steady-state and the manual work will be wrapped up.

files/Makefile
```diff
@@ -28,58 +28,12 @@
...
-	-e TARGET_ARCH="$(TARGET_ARCH)" \
-	-e TARGET_OS="$(TARGET_OS)" \
-	-v /etc/passwd:/etc/passwd:ro \
-	$(DOCKER_SOCKET_MOUNT) \
-	$(CONTAINER_OPTIONS) \
-	--mount type=bind,source="$(PWD)",destination="/work" \
-	--mount type=bind,source="$(TARGET_OUT)",destination="/targetout" \
-	--mount type=volume,source=home,destination="/home" \
-	-w /work $(IMG)
+$(info Building with the build container: $(IMG), outputting to $(TARGET_OUT).)
```
sdake: can you make these separate lines?  The image name is long, and the path is long, which results in wrapping.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,79 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# set -ex
+
+local_arch=$(uname -m)
```
sdake: we want to be able to override TARGET_ARCH from the makefile.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,79 @@
...
+elif [[ $local_arch == armv8* ]]; then
+    target_arch=arm64
+elif [[ $local_arch == armv* ]]; then
+    target_arch=arm
+else
+    echo "This system's architecture, $local_arch, isn't recognized/supported"
+    exit 1
+fi
+
+local_os=$(uname)
```
sdake: same thing with TARGET_OS

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,79 @@
...
+
+timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
+
+repo_root=$(git rev-parse --show-toplevel)
+repo_name=$(basename "$repo_root")
+target_out="${TARGET_OUT:-$HOME/istio_out/$repo_name}"
+
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+
+container_cli="${CONTAINER_CLI:-docker}"
```
sdake: how does CONTAINER_CLI make it into this script from the makefile?

files/Makefile
XanderStrike: Good call. Since `TARGET_OUT` is relevant whether it's in a container or not I'll bump that line out of the `ifeq` as well.

files/common/scripts/run-docker.sh
XanderStrike: As above https://github.com/istio/common-files/pull/94/files#r332134900

files/common/scripts/run-docker.sh
XanderStrike: As above https://github.com/istio/common-files/pull/94/files#r332134900

files/common/scripts/run-docker.sh
XanderStrike: I'll create a new issue for passing arbitrary environment variables.

geeknoid:
Note that as you continue iterating on this PR, the underlying makefile-based stuff has been evolving a bit. So you'll need to reconcile all that before submitting the final PR.

Thanks.

XanderStrike:
The changes to the makefile haven't been substantial so far. Let me know when we're ready to merge this PR and I'll address the conflicts.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+# $CONTAINER_OPTIONS becomes an empty arg when quoted, so SC2086 is disabled for the
+# following command only
+# shellcheck disable=SC2086
+$container_cli run -it --rm -u "$(id -u):docker" \
+    --sig-proxy=true \
+    ${DOCKER_SOCKET_MOUNT:+"-v /var/run/docker.sock:/var/run/docker.sock"} \
+    -v /etc/passwd:/etc/passwd:ro \
+    $CONTAINER_OPTIONS \
+    -e IN_BUILD_CONTAINER=1 \
+    -e TZ="${timezone:-$TZ}" \
```
sdake: @XanderStrike you didn't create this problem - so you don't necessarily have to fix it... 

Several of our build scripts operate on external environment variables.  If the environment variables re not passed via `-e`, then the makefiles we already have will not always function properly.  See: https://github.com/istio/common-files/pull/106 for an example.  I don't want to keep adding `-e` flags for each variable as we find them.  Could you sort out a mechanism whereby `-e` flags are autogenerated based upon the existing environment, but ignoring certain systemwide environment variables? (`PATH` would be a prime example of something to ignore in the creation of this list).

As we are in the final stretch of 1.4, I'm not sure we can introduce this work atleast until we branch, and possibly after that as we don't want to disrupt developer workflows.  IOW, you have some time :)

BTW, existing work looks great.

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
+
+container_cli="${CONTAINER_CLI:-docker}"
+
+$container_cli pull "$image"
+
+# $CONTAINER_OPTIONS becomes an empty arg when quoted, so SC2086 is disabled for the
+# following command only
+# shellcheck disable=SC2086
+$container_cli run -it --rm -u "$(id -u):docker" \
```
sdake: even though the shellcheck is disabled, it would be a good idea to quote all shell calls.  e.g. `"${container_cli}"`.

files/common/scripts/run-docker.sh
sdake: @XanderStrike i'd encourage working on this now if you have free cycles.  It may be some time until this PR merges, but it will merge (at-least i'd merge it).

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+else
+    echo "This system's OS, $local_os, isn't supported"
+    exit 1
+fi
+
+timezone=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
+
+out="${TARGET_OUT:$(pwd)/out/${target_arch}_${target_os}}"
+
+image="${IMG:-gcr.io/istio-testing/build-tools:latest}"
```
howardjohn: I don't think we ever want this to fall out of sync with the IMG from the makefile, otherwise 2 images are downloaded and they may behave differently. Do we need the IMG in the makefile, or can we just make the source of truth in this file?

files/common/scripts/run-docker.sh
```diff
@@ -0,0 +1,77 @@
...
+
+container_cli="${CONTAINER_CLI:-docker}"
+
+$container_cli pull "$image"
```
howardjohn: nit: is there any benefit to doing `pull && run` instead of just `run`? it will be pulled if needed during run

files/common/scripts/run-docker.sh
howardjohn: I don't think this disrupts dev workflows, this should be transparent to anyone that isn't paying close attention, right?

sdake: if you mean this PR, as long as it works :)

files/common/scripts/run-docker.sh
XanderStrike: Good catch, thanks!

files/common/scripts/run-docker.sh
XanderStrike: I initially left it because I didn't want whatever was updating it to have to change their process. Now that I know it's a manual process I don't mind moving it into the script.

files/common/scripts/run-docker.sh
XanderStrike: If you use the `latest` tag (or any tag you regularly push new images to), `docker run` will only check in with the remote repository if there's not already an image tagged with `latest` locally. The alternative of `docker run --pull=always` isn't great because it'll literally always pull, so `docker pull && docker run` is the best way of dealing with this IMO.

Of course, we're not using the `latest` tag [any more](https://github.com/istio/common-files/pull/94#discussion_r334305815) so I could remove it, but it doesn't cost anything to run it every time and supports the possibility that a consumer of this script might want to use a less specific tag for.

files/common/scripts/run-docker.sh
XanderStrike: Fair enough, I've added an initial implementation now, let me know what you think

files/common/scripts/run-docker.sh
XanderStrike: The host's env is now string-crunched into a list of `-e ` flags (with some filtering to prevent if from screwing up the internal environment).

howardjohn:
@sdake @geeknoid what is blocking this change? Can you please leave comments of what is missing and I will address them and we can merge this in?

sdake:
@howardjohn the only thing blocking this work is the 1.4 release IMO.

We are getting close to meeting that objective, so I think we can merge this work soon once it goes through another review.

I spoke with @XanderStrike via slack at length about our plans for common-files, and he seemed ok with the approach of waiting until we have reached a stable rollout of all of the feature set.  We are very close to reaching this point - waiting only on docker for desktop `make docker.push` work to finish out monday or tuesday.

Cheers
-steve

sdake:
@XanderStrike 1.4 has released yesterday.  If you can rev this to the latest implementation of the Makefile, I think we can get it merged relatively quickly.  There are alot of special cases in the Makefile and sadly only a few of them are documented.  I will be at kubecon next week and not super available to review this PR.  After Kubecon, we should make a priority of simplifying the top-level makefile with this shell script.   If you will be at Kubecon, perhaps we can meet :)

Cheers
-steve

howardjohn:
replaced by https://github.com/istio/common-files/pull/172

PR #97, opened 2019-10-05
Initialize docker for DIND/docker environments

sdake:


sdake:
@howardjohn https://github.com/istio/cni/pull/191

sdake:
@howardjohn thiis will be needed (just as it is needed in prow) in any kind use cases involving e2e testing).  In prow, this already happens.  locally it doesn't.

sdake:
replacing with https://github.com/istio/common-files/pull/98

PR #100, opened 2019-10-09
Set TARGET_OUT within container properly

sdake:


files/Makefile
```diff
@@ -53,6 +53,7 @@ endif
 export TARGET_OUT ?= $(shell pwd)/out/$(TARGET_ARCH)_$(TARGET_OS)
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
+export TARGET_OUT = /work/out/$(TARGET_ARCH)_$(TARGET_OS)
```
XanderStrike: This makes it impossible to override TARGET_OUT, was this intentional?

geeknoid: I don't see any value in overriding this value for artifacts within the container, but for the sake of uniformity, it would be better to allow it to be overriden. I'd turn this into ?= and move line 53 into the else clause below.

PR #102, opened 2019-10-10
update to latest container image

jasonwzm:


geeknoid:
I'm going to have another update in a bit, a container is building at the moment...

jasonwzm:
oh my bad... then i will wait for the new image.

PR #109, opened 2019-10-18
Makefile updates.

geeknoid:
- Use latest container image.

- Make the branch fetched by update-common and update-common-protos 
be configurable.

- Make update-common and update-common-protos use a temp directory to clone the
common-files repo so as to not temporarily pollute the destination repo and confuse
my IDE in the process.

- Simplify how the non-container case is handled. Instead of recursively
invoking make, just include Makefile.core.mk.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -80,19 +80,24 @@ dump-licenses-csv:
 	@go mod download
 	@license-lint --config common/config/license-lint.yml --csv
 
+TMP := $(shell mktemp -d -u)
+UPDATE_BRANCH ?= "master"
+
 update-common:
-	@git clone -q --depth 1 --single-branch --branch master https://github.com/istio/common-files
-	@cd common-files ; git rev-parse HEAD >files/common/.commonfiles.sha
+	@mkdir -p $(TMP)
+	@git clone -q --depth 1 --single-branch --branch $(UPDATE_BRANCH) https://github.com/istio/common-files $(TMP)/common-files
+	@cd $(TMP)/common-files ; git rev-parse HEAD >files/common/.commonfiles.sha
 	@rm -fr common
-	@cp -rT common-files/files .
-	@rm -fr common-files
+	@cp -rT $(TMP)/common-files/files $(pwd)
+	@rm -fr $(TMP)/common-files
 
 update-common-protos:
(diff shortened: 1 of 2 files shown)


files/common/Makefile.common.mk
```diff
@@ -80,19 +80,23 @@ dump-licenses-csv:
 	@go mod download
 	@license-lint --config common/config/license-lint.yml --csv
 
+TMP := $(shell mktemp -d -u)
```
sdake: -u is unsafe - can present a race coditioin.

files/common/Makefile.common.mk
```diff
@@ -80,19 +80,23 @@ dump-licenses-csv:
...
+
 update-common:
-	@git clone -q --depth 1 --single-branch --branch master https://github.com/istio/common-files
-	@cd common-files ; git rev-parse HEAD >files/common/.commonfiles.sha
+	@mkdir -p $(TMP)
```
sdake: why not try TMP = $(shell mktemp) here?  Alternatively you could make it a target, which should set it globally if you use `export $(TMP)`

geeknoid: I've been unable to assign a variable within a recipe, it comes back empty when I read it back.

files/common/Makefile.common.mk
geeknoid: Yes, if somehow folks run the same container and run the same make target in that container at the same time, then it'll be racy. Not sure I care though :-)

sdake: but `update-common` target isn't run in the container?

files/common/Makefile.common.mk
sdake: did you try `export $(TMP)`?

geeknoid:
export $(TMP) doesn't work.

Yes, I realize the update-common wouldn't work with people having random branches. The magic file might work, but I think I've got a simpler idea. I'll have an update shortly.

And finally, update-common definitely does run from within the contain. In fact, for Mac users, it can't be used outside the container at the moment since it uses some options to the cp command (if I recall correctly) which aren't available on macOS.

files/common/Makefile.common.mk
geeknoid: Doesn't work...

PR #110, opened 2019-10-21
Fix TARGET_OUT to match istio/istio

howardjohn:
See https://github.com/istio/istio/blob/89a6afc32ba3fd3a94776875c7eaf0650dfcdb32/Makefile.core.mk#L141


howardjohn:
@sdake operator will not be published until this is done

PR #112, opened 2019-10-22
Add HUB and ISTIO_VERSION variables for build istio/operator

carolynhu:
https://github.com/istio/istio/pull/18185
https://github.com/istio/istio/issues/18173

### DIFF
--- a/files/common/scripts/report_build_info.sh
+++ b/files/common/scripts/report_build_info.sh
@@ -43,9 +43,11 @@ if [[ -n ${ISTIO_VERSION} ]]; then
 fi
 
 GIT_DESCRIBE_TAG=$(git describe --tags)
+HUB=${HUB:-""}
 
 # used by common/scripts/gobuild.sh
 echo "istio.io/pkg/version.buildVersion=${VERSION}"
 echo "istio.io/pkg/version.buildGitRevision=${BUILD_GIT_REVISION}"
 echo "istio.io/pkg/version.buildStatus=${tree_status}"
 echo "istio.io/pkg/version.buildTag=${GIT_DESCRIBE_TAG}"
+echo "istio.io/pkg/version.buildHub=${HUB}"


files/common/scripts/report_build_info.sh
```diff
@@ -43,9 +43,12 @@ if [[ -n ${ISTIO_VERSION} ]]; then
...
 
 GIT_DESCRIBE_TAG=$(git describe --tags)
+HUB=${HUB:-""}
 
 # used by common/scripts/gobuild.sh
 echo "istio.io/pkg/version.buildVersion=${VERSION}"
 echo "istio.io/pkg/version.buildGitRevision=${BUILD_GIT_REVISION}"
 echo "istio.io/pkg/version.buildStatus=${tree_status}"
 echo "istio.io/pkg/version.buildTag=${GIT_DESCRIBE_TAG}"
+echo "istio.io/operator/cmd/mesh.HubValueFromBuild=${HUB}"
```
geeknoid: This doesn't feel right. This file is used in different repos, so having operator-specific definitions here is probably inappropriate.

@howardjohn

howardjohn: I considered this as well, but then we will need a fork of this for istio/istio. That seems worse than this?

geeknoid: I guess I don't understand what makes the operator different. Isn't there a similar issue with cni?

howardjohn: CNI is just a go binary, no different from pilot or galley. It uses the above pkg/version... stuff to allow `cni --version` to work. For the hub and tag, we rewrite those at release time in the release yamls. The operator is different, it embeds those yamls into the actual binary with vfs gen. These yamls (really go code, but embedded yaml) need to be overwritten at release time with the correct hub and tag.

geeknoid:
When the operator runs, it can find its own version from the pkg/version API. Given this, it could trivially stamp its embedded YAML dynamically without involving the build system to have custom logic or other custom shenanigans. Or am I missing something?

ostromart:
@geeknoid it could do tag but not hub.

howardjohn:
+1 to what @ostromart said. Even if we did have all the info, `istio.io/pkg/version.buildTag` is a private variable (maybe it exposes it through some other method though)

geeknoid:
Then let's extend pkg/version.BuildInfo to carry the hub and tag alongside the other information it already does, and then this change will be meaningful across the codebase.

howardjohn:
We used to have the docker hub as part of the build info and removed it
because it doesn't make sense for a pilot or galley binary to have a docker
hub - it's a golang binary.

So we can add it to Istio/pkg as long as it doesn't show up in --version

geeknoid:
Works for me. These values can be in the returned BuildInfo and read by
operator and not displayed normally.

kyessenov:
FYI I am adding docker hub to proxy build. It makes the push reproducible.

howardjohn:
@kyessenov what do you mean, the docker hub will be embedded somewhere in the proxy binary? What is the reasoning for it?

kyessenov:
Bazel requires all inputs know prior to the build so that everything is done exactly the same. Since we are adding docker push, and the repository is part of that operation, the build needs to know the repo ahead of time. That way when you run `bazel run`, it will never corrupt or override the pushed image.

carolynhu:
istio/pkg PR: https://github.com/istio/pkg/pull/81

PR #117, opened 2019-10-24
Common file updates.

geeknoid:
- Use latest container image.

- Added the tidy-go target, and made format-go depend on it.

- Fixed an error message about updating generated content.


geeknoid:
I'm going to do the rounds today and update common files everywhere. So
I'll include the golang update.

PR #118, opened 2019-10-24
Update golangci-lint version to match what's in the container.

geeknoid:


howardjohn:
and remove GOGC=25 from lint_go.sh, this takes 1/4 as much memory as it used to. I tested on istio/istio with it off and it peaked at 4gb which is reasonable even for a laptop

geeknoid:
PTAL

howardjohn:
Changes required on istio/istio to get this passing: https://github.com/istio/istio/pull/18279

PR #122, opened 2019-10-25
Port a few common file changes from master.

geeknoid:
- And use latest 1.4 build container.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -56,7 +56,7 @@ ifeq ($(BUILD_WITH_CONTAINER),1)
 export TARGET_OUT = /work/out/$(TARGET_ARCH)_$(TARGET_OS)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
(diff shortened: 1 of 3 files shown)


files/Makefile
```diff
@@ -56,7 +56,7 @@ ifeq ($(BUILD_WITH_CONTAINER),1)
 export TARGET_OUT = /work/out/$(TARGET_ARCH)_$(TARGET_OS)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
-IMG ?= gcr.io/istio-testing/build-tools:2019-10-11T13-37-52
+IMG ?= gcr.io/istio-testing/build-tools:2019-10-25T16-21-08
```
howardjohn: I think this should be from the 1.4 branch?

geeknoid: It is, it's an image built from the 1.4 version of the Dockerfile.

Our image labeling isn't helping this....

geeknoid: See https://prow.istio.io/view/gcs/istio-prow/logs/tools-containers_release-1.4/7

PR #145, opened 2019-11-12
Add option to use proxy builder image

bianpengyuan:
Change build to use proxy build-tools when path basename is `proxy`. Also add USER env variable to docker run, which is needed by proxy build.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -57,6 +57,9 @@ export TARGET_OUT = /work/out/$(TARGET_OS)_$(TARGET_ARCH)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
 IMG ?= gcr.io/istio-testing/build-tools:master-2019-11-08T17-21-18
+ifeq ($(shell basename $(shell pwd)),proxy)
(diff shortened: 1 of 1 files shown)


files/Makefile
```diff
@@ -57,6 +57,9 @@ export TARGET_OUT = /work/out/$(TARGET_OS)_$(TARGET_ARCH)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
 IMG ?= gcr.io/istio-testing/build-tools:master-2019-11-08T17-21-18
+ifeq ($(shell basename $(shell pwd)),proxy)
+IMG = gcr.io/istio-testing/build-tools-proxy:master-2019-11-08T17-21-18
```
howardjohn: @geeknoid can you override the IMG in Makefile.override.mk?

That may be a better solution than hard coding this to proxy, although this should work too

geeknoid: Yeah, the intent was to override IMG in the proxy repo.

bianpengyuan: I see. Thanks! Will use the override file instead.

howardjohn:
Do you still need the USER? Not sure if that can be added in overrides

bianpengyuan:
@howardjohn Yeah, realized that after closing the pr :). Removed proxy image selection logic.

PR #149, opened 2019-11-15
Make lint-copyright-banner ignore paths in FILTER_COPYRIGHT_WHITELIST

skaslev:
kubetype-gen generated `doc.go` files have no copyright header. Before istio/client-go#22 they got renamed to `doc.gen.go` and hence ignored by the `lint-copyright-banner` target. 

Since k8s code generator expects it to be called exactly `doc.go` in order to parse the api `groupName`, ignore them in `lint-copyright-banner`.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -22,6 +22,13 @@
 FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./.github -o -path ./licenses \) -prune -o -type f
 XARGS = xargs -0 -r
 
+comma := ,
+ifeq ($(COPYRIGHT_LINT_WHITELIST),)
+  FILTER_COPYRIGHT_WHITELIST=cat
+else
+  FILTER_COPYRIGHT_WHITELIST=egrep -vxzZ -e ./$(subst $(comma),|./,$(COPYRIGHT_LINT_WHITELIST))
+endif
+
 lint-dockerfiles:
 	@${FINDFILES} -name 'Dockerfile*' -print0 | ${XARGS} hadolint -c ./common/config/.hadolint.yml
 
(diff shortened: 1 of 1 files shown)


geeknoid:
I'd rather not have repo-specific idiosyncrasies in the common files. Instead, how about adding a variable to let repos inject exceptions? As an example, we use the MARKDOWN_LINT_WHITELIST variable which can be defined in a repo's Makefile.overrides.mk file. Similarly, we can introduce COPYRIGHT_LINT_WHITELIST and have it set appropriately in client-go's Makefile.overrides.mk.

WDYT?

nrjpoddar:
> I'd rather not have repo-specific idiosyncrasies in the common files. Instead, how about adding a variable to let repos inject exceptions? As an example, we use the MARKDOWN_LINT_WHITELIST variable which can be defined in a repo's Makefile.overrides.mk file. Similarly, we can introduce COPYRIGHT_LINT_WHITELIST and have it set appropriately in client-go's Makefile.overrides.mk.
> 
> WDYT?

Makes perfect sense, however if this enhancement turns out to be bigger than expected, I would recommend opening an issue and merging this in, so we can track work separately and not block this PR.

skaslev:
@geeknoid I added `COPYRIGHT_LINT_WHITELIST` which will get ignored by lint-copyright-banner. 

`COPYRIGHT_LINT_WHITELIST` is being set in client-go's `Makefile.core.mk` to
```
COPYRIGHT_LINT_WHITELIST:='pkg/apis/.*/doc.go,pkg/clientset/.*/doc.go,pkg/informers/.*/doc.go,pkg/listers/.*/doc.go'
```

Can you take another look? Criticism is very welcome since my bash/make fu is not very strong :)

sdake:
@skaslev hate to jump in here at the last minute, but what licenses do you want to have the linter ignore?  I guess I don't understand the purpose of this PR.  Could you provide a little more context?  All licenses used in anything we use should go through a review process, and I'm concerned this may be a way around that.

Cheers
-steve

sdake:
@skaslev ignore last comment - I see that the description issue was edited from the first time I reviewed this PR.  One question I have is why can't the licenses be autogenerated from the code?  And under what license is this generated code?

I've put a hold on this PR until the question is answered - @geeknoid feel free to override my hold if you're satisfied with the legal aspects.

I do have some minimal concerns that this may result in license non-compliance.  The implications of this PR require all our reviewers unfamiliar with the licensing issues to become experts in escalation to the attorneys when this whitelist is in use in the repo in question.

Cheers
-steve

skaslev:
@sdake Sure, we can just add copyright notice to all istio k8s client-go generated files instead. I'm not sure why this wasn't the case in the first place.

Closing this PR.

sdake:
@skaslev great - thanks for doing the work there.  Hopefully that is legally ok.  Its hard to say why this was not the case in the past, but I'm good with the coice :)

Cheers
-steve

PR #152, opened 2019-11-21
Fix Makefile bugs on Arm64 platform

Jingzhao123:
The command `uname -m` on arm64 platform return the
value 'aarch64', but not arm64.

Signed-off-by: Jingzhao <Jingzhao.Ni@arm.com>

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -31,7 +31,7 @@ export BUILD_WITH_CONTAINER ?= 0
 LOCAL_ARCH := $(shell uname -m)
 ifeq ($(LOCAL_ARCH),x86_64)
     TARGET_ARCH ?= amd64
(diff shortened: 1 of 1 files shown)


Jingzhao123:
@googlebot I signed it!

files/Makefile
```diff
@@ -31,7 +31,7 @@ export BUILD_WITH_CONTAINER ?= 0
 LOCAL_ARCH := $(shell uname -m)
 ifeq ($(LOCAL_ARCH),x86_64)
     TARGET_ARCH ?= amd64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
+else ifeq ($(LOCAL_ARCH),aarch64)
```
howardjohn: (I am not familiar with this area) is this breaking `armv8`? Should we set this if the response is `armv8` OR `aarch64`?

PR #162, opened 2019-12-04
update container image

jasonwzm:


howardjohn:
I think this will break any repo with go code, since we downgraded the linter and need to update the linter config. Let me send a fix for that

jasonwzm:
Close in favor of https://github.com/istio/common-files/pull/163

PR #167, opened 2019-12-07
Add istio api extensions to common-files

jasonwzm:


jasonwzm:
@sdake @geeknoid could you please take a look?

jasonwzm:
@sdake so are you saying we create a `protos/` directory in this repo and put extensions proto in it? And istio/api would also have a `protos/` directory which is copied from this repo?

sdake:
> @sdake so are you saying we create a `protos/` directory in this repo and put extensions proto in it? And istio/api would also have a `protos/` directory which is copied from this repo?

not quite.. `isito/api` would continue to have a common-protos directory.  This repo would contain `/protos` which would then be synced to `common-protos` in this repo.

PR #168, opened 2019-12-12
Add conditional host mount of $HOME/.kube

sdake:


files/Makefile
```diff
@@ -73,14 +73,21 @@ TIMEZONE=`readlink $(READLINK_FLAGS) /etc/localtime | sed -e 's/^.*zoneinfo\///'
...
 endif
+
 ifneq (,$(wildcard $(HOME)/.config/gcloud))
 $(info Using gcr credential directory $(HOME)/.config/gcloud.)
-DOCKER_CREDS_MOUNT+=--mount type=bind,source="$(HOME)/.config/gcloud",destination="/config/.config/gcloud",readonly
+CONDITIONAL_HOST_MOUNTS+=--mount type=bind,source="$(HOME)/.config/gcloud",destination="/config/.config/gcloud",readonly
+endif
+
+ifneq (,$(wildcard $(HOME)/.kube))
+$(info Using local Kubernetes configuration $(HOME)/.kube)
```
howardjohn: nit: can we make these debug so they only show up with `-v` or something? not sure if make supports this

sdake: sure - we can make these debug or turn them off or remove them if people prefer. They are not there for debug purposes, they are there to tell developers what stuff is being imported from their environment (as these are all security credentials).

PR #172, opened 2019-12-23
Add docker run script

howardjohn:
Take over https://github.com/istio/common-files/pull/94/files

This makes it possible to run interactive commands inside the build container

resolves istio/istio#17476 - all commands arguments to run-docker.sh are passed
into the container

resolves istio/istio#17474 - the script will pass along a filtered picture of the host's environment to the container

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -30,8 +30,19 @@ SHELL := /bin/bash
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
+# Set up environment variables
+$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
+include out/.env
+export out/.env
+
 ifeq ($(BUILD_WITH_CONTAINER),1)
+
+# Override variables with container specific
+export TARGET_OUT=$(CONTAINER_TARGET_OUT)
+export TARGET_OUT_LINUX=$(CONTAINER_TARGET_OUT_LINUX)
+
 RUN = ./common/scripts/run.sh
+IMG ?= gcr.io/istio-testing/build-tools:master-2019-12-15T16-17-48
 
 # If these variables are defined in the overrides, exporting them adds them to
 # the environment of the sub-make
@@ -44,6 +55,8 @@ export CONTAINER_OPTIONS
 export TZ
 export HUB
 export TAG
+export RUN
+export IMG
 
 MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
 
--- a/files/Makefile
+++ b/files/Makefile
@@ -30,34 +30,13 @@ SHELL := /bin/bash
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
-# Set up environment variables
+# Set up environment variables. Using a common script allows the script and makefile to re-use logic
 $(shell mkdir -p out)
-$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
-include out/.env
-export out/.env
+$(shell $(PWD)/common/scripts/setup_env.sh)
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
 
-# Override variables with container specific
-export TARGET_OUT=$(CONTAINER_TARGET_OUT)
-export TARGET_OUT_LINUX=$(CONTAINER_TARGET_OUT_LINUX)
-
 RUN = ./common/scripts/run.sh
-IMG ?= gcr.io/istio-testing/build-tools:master-2019-12-15T16-17-48
-
-# If these variables are defined in the overrides, exporting them adds them to
-# the environment of the sub-make
-export TARGET_ARCH
-export TARGET_OS
-export TARGET_OUT
-export CONTAINER_CLI
-export DOCKER_SOCKET_MOUNT
-export CONTAINER_OPTIONS
-export TZ
-export HUB
-export TAG
-export RUN
-export IMG
 
 MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
 
--- a/files/Makefile
+++ b/files/Makefile
@@ -31,6 +31,7 @@ SHELL := /bin/bash
 export BUILD_WITH_CONTAINER ?= 0
 
 # Set up environment variables
+$(shell mkdir -p out)
 $(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
 include out/.env
 export out/.env
--- a/files/Makefile
+++ b/files/Makefile
@@ -30,100 +30,20 @@ SHELL := /bin/bash
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
-# Version of image used within build container
-IMAGE_VERSION ?= master-2020-03-05T18-27-04
-
-LOCAL_ARCH := $(shell uname -m)
-ifeq ($(LOCAL_ARCH),x86_64)
-    TARGET_ARCH ?= amd64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
-    TARGET_ARCH ?= arm64
-else ifeq ($(LOCAL_ARCH),aarch64)
-    TARGET_ARCH ?= arm64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 4),armv)
-    TARGET_ARCH ?= arm
-else
-    $(error This system's architecture $(LOCAL_ARCH) isn't supported)
-endif
-
-LOCAL_OS := $(shell uname)
-ifeq ($(LOCAL_OS),Linux)
-    TARGET_OS ?= linux
-    READLINK_FLAGS="-f"
-else ifeq ($(LOCAL_OS),Darwin)
-    TARGET_OS ?= darwin
-    READLINK_FLAGS=""
-else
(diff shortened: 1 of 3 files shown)


howardjohn:
Ok I split out the enviroment variables, should be ok now

files/Makefile
```diff
@@ -28,95 +28,34 @@
...
-else
-    $(error This system's OS $(LOCAL_OS) isn't supported)
-endif
-
-export TARGET_OUT ?= $(shell pwd)/out/$(TARGET_OS)_$(TARGET_ARCH)
-export TARGET_OUT_LINUX ?= $(shell pwd)/out/linux_amd64
+# Set up environment variables
+$(shell mkdir -p out)
+$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
+include out/.env
```
sdake: I don't understand this include. Why not simply pass the variables in this file? I am not keen on using filesystem contents in the top level Makefile.

files/Makefile
```diff
@@ -28,95 +28,34 @@
...
-
-ENV_VARS:=
-ifdef HUB
-ENV_VARS+=-e HUB="$(HUB)"
-endif
-ifdef TAG
-ENV_VARS+=-e TAG="$(TAG)"
-endif
+RUN = ./common/scripts/run.sh
+IMG ?= gcr.io/istio-testing/build-tools:master-2019-12-15T16-17-48
```
sdake: IMG version needs updating.

files/Makefile
```diff
@@ -28,95 +28,34 @@
...
-
-$(info Building with the build container: $(IMG).)
 
-# Determine the timezone across various platforms to pass into the
-# docker run operation. This operation assumes zoneinfo is within
-# the path of the file.
-TIMEZONE=`readlink $(READLINK_FLAGS) /etc/localtime | sed -e 's/^.*zoneinfo\///'`
+# Override variables with container specific
+export TARGET_OUT=$(CONTAINER_TARGET_OUT)
+export TARGET_OUT_LINUX=$(CONTAINER_TARGET_OUT_LINUX)
```
sdake: The introduction of CONTAINER_* variables seems confusing to me?

files/common/scripts/run.sh
```diff
@@ -0,0 +1,53 @@
...
+# following command only
+# shellcheck disable=SC2086
+"${CONTAINER_CLI}" run -it --rm \
+    -u "${UID}:${DOCKER_GID}" \
+    --sig-proxy=true \
+    ${DOCKER_SOCKET_MOUNT:+"-v /var/run/docker.sock:/var/run/docker.sock"} \
+    -v /etc/passwd:/etc/passwd:ro \
+    -v /etc/group:/etc/group:ro \
+    $CONTAINER_OPTIONS \
+    --env-file <(env | grep -v ${ENV_BLACKLIST}) \
```
sdake: nice simple approach here).

files/common/scripts/run.sh
```diff
@@ -0,0 +1,53 @@
...
+    -e IN_BUILD_CONTAINER=1 \
+    -e TZ="${TIMEZONE:-$TZ}" \
+    --mount "type=bind,source=${PWD},destination=/work" \
+    --mount "type=volume,source=go,destination=/go" \
+    --mount "type=volume,source=gocache,destination=/gocache" \
+    ${CONDITIONAL_HOST_MOUNTS} \
```
sdake: default CONDITIOANL_HOST_MOUNTS missing.

files/common/scripts/setup_env.sh
```diff
@@ -0,0 +1,98 @@
...
+    readlink_flags="-f"
+elif [[ $LOCAL_OS == Darwin ]]; then
+    export TARGET_OS=darwin
+    readlink_flags=""
+else
+    echo "This system's OS, $LOCAL_OS, isn't supported"
+    exit 1
+fi
+
+export UID
```
sdake: where is UID set? Not in the environment:

```
sdake@beast-01:~/go/src/istio.io/istio$ env | grep UID
```

files/common/scripts/setup_env.sh
```diff
@@ -0,0 +1,98 @@
...
+TIMEZONE=$(readlink $readlink_flags /etc/localtime | sed -e 's/^.*zoneinfo\///')
+export TIMEZONE
+
+export TARGET_OUT="${TARGET_OUT:-${PWD}/out/${TARGET_OS}_${TARGET_ARCH}}"
+export TARGET_OUT_LINUX="${TARGET_OUT_LINUX:-${PWD}/out/linux_amd64}"
+
+export CONTAINER_TARGET_OUT="${CONTAINER_TARGET_OUT:-/work/out/${TARGET_OS}_${TARGET_ARCH}}"
+export CONTAINER_TARGET_OUT_LINUX="${CONTAINER_TARGET_OUT_LINUX:-/work/out/linux_amd64}"
+
+export IMG="${IMG:-gcr.io/istio-testing/build-tools:master-2019-12-15T16-17-48}"
```
sdake: would be nice if we had one IMG variable to modify rather than two.

files/common/scripts/setup_env.sh
```diff
@@ -0,0 +1,98 @@
...
+
+export CONTAINER_CLI="${CONTAINER_CLI:-docker}"
+
+export ENV_BLACKLIST="${ENV_BLACKLIST:-^_\|PATH\|SHELL\|EDITOR\|TMUX\|USER\|HOME\|PWD\|TERM\|GO\|rvm\|SSH}"
```
sdake: call this ENV_REJECTLIST please.

files/Makefile
howardjohn: We need a way to have the environment setting consistent across the makefile and the shell script. This was the only way I could come up with this, but if there are better approaches I am all ears, I didn't feel great about this

files/Makefile
howardjohn: The whole PR is out of date a bit, I'll rebase once we are ready to merge

files/common/scripts/run.sh
howardjohn: Its in the environment setup script

files/common/scripts/setup_env.sh
howardjohn: Oh maybe this isn't right then. UID is a builtin I guess

```
$ env | grep UID
$ echo $UID
576696
```

files/Makefile
howardjohn: I am not sure why I have this IMG, it should just be defined in the script. Ill see if I can remove this one

files/common/scripts/setup_env.sh
howardjohn: ack, I think the other one can be removed

files/Makefile
howardjohn: The intent here was that we are defining thins like TARGET_OUT conditionally in the Makefile so we need to have both defined in the script -- no one other than Makefile uses the container_out though.

However, we can probably do the BUILD_WITH_CONTAINER check inside the script an remove the need for this

files/Makefile
sdake: I don't have any immediate ideas, but I am beat after 5 hours of popcorn ceiling removal. Wow what a PITA. We can always adjust this in the future.

files/common/scripts/setup_env.sh
sdake: I'd verify the export works as expected. I would think it should as a built-in but best to be sure.

sdake:
Understood re difficulty of rebasing - especially against a moving target. I think BUILD_WITH_CONTAINER=1 PR is ready to merge. Once that is done, I'd suggest moving on this fast before more changes come to the TLD makefile in the common-files repo 😁

sdake:
@howardjohn what do you want to do with this PR post 1.5 release? I think the make system is in sold enough shape that a script based approach can be used. Best to do that at the beginning of the dev cycle (after 1.5 work is done of course).

Your thoughts welcome as I am listed as a reviewer.

howardjohn:
I would like to merge this in a similar form as it is today but have not
yet found the time to do so

howardjohn:
If anyone else is interested feel free to take this over

howardjohn:
I did not mean to merge this, I didn't realize it was approved. However, I do think it should be merge. If anyone disagrees we can revert or roll forward?

clarketm:
> I did not mean to merge this, I didn't realize it was approved. However, I do think it should be merge. If anyone disagrees we can revert or roll forward?

There is probably an assumption that this is to run on Linux, but FWIW `make` commands no longer work for me locally on MacOS.

```console
$ BUILD_WITH_CONTAINER=1 make lint

/bin/sh: 1: set: Illegal option -o pipefail
common/Makefile.common.mk:33: recipe for target 'lint-yaml' failed
make[1]: *** [lint-yaml] Error 2
make: *** [Makefile:44: lint] Error 2
```

```console
$ BUILD_WITH_CONTAINER=1 make update-common

mkdir: cannot create directory '/var/folders': Permission denied
common/Makefile.common.mk:98: recipe for target 'update-common' failed
make[1]: *** [update-common] Error 1
make: *** [Makefile:44: update-common] Error 2
```

```console
$ BUILD_WITH_CONTAINER=1 make test

go: creating work dir: stat /var/folders/9j/nh3fxdf153v11vsjwtd1lf_800n1b6/T/: no such file or directory
Makefile.core.mk:23: recipe for target 'test' failed
make[1]: *** [test] Error 1
make: *** [Makefile:44: test] Error 2
```

PR #173, opened 2019-12-26
Fixed the Istio docker build failure

seflerZ:
This PR includes:
✔ Fixed corrupted docker building procedure in Istio component.
✔ Enable "go test" options by setting environment variable $T which is already existed in "Makefile.core.mk"

I'll explain this in details:
👉 I've made the "current directory" mapped to  "/work/src/istio.io/istio" because some legacy tests in Istio need path "/src/istio.io/istio" to find the source root. This doesn't affect the final outputs and behaves exactly like "building with local tool chain".
👉 Added the missing "GOBIN" environment variable (Otherwise it will become "/gobin" in building process which will cause test failure).

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -108,17 +108,15 @@ RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):$(GID) --rm \
 	-e TARGET_OUT="$(TARGET_OUT)" \
 	-e TARGET_OUT_LINUX="$(TARGET_OUT_LINUX)" \
 	-e USER="${USER}" \
-	-e GOPATH="/work" \
-	-e GOBIN="/work/bin" \
 	$(ENV_VARS) \
 	-v /etc/passwd:/etc/passwd:ro \
 	$(DOCKER_SOCKET_MOUNT) \
 	$(CONTAINER_OPTIONS) \
-	--mount type=bind,source="$(GOPATH)",destination="/work" \
+	--mount type=bind,source="$(PWD)",destination="/work" \
 	--mount type=volume,source=go,destination="/go" \
 	--mount type=volume,source=gocache,destination="/gocache" \
 	$(CONDITIONAL_HOST_MOUNTS) \
-	-w /work/src/istio.io/istio $(IMG)
+	-w /work $(IMG)
 
 MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
 
--- a/files/Makefile
+++ b/files/Makefile
@@ -28,7 +28,7 @@
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
-# This is the "go test" options
+# This is the "go test" options.
 export T ?= "--timeout=100s"
 
 LOCAL_ARCH := $(shell uname -m)
--- a/files/Makefile
+++ b/files/Makefile
@@ -29,7 +29,7 @@
 export BUILD_WITH_CONTAINER ?= 0
 
 # This is the "go test" options.
-export T ?= "--timeout=100s"
+export T ?= ""
 
 LOCAL_ARCH := $(shell uname -m)
 ifeq ($(LOCAL_ARCH),x86_64)
--- a/files/Makefile
+++ b/files/Makefile
@@ -59,8 +59,8 @@ export TARGET_OUT ?= $(shell pwd)/out/$(TARGET_OS)_$(TARGET_ARCH)
 export TARGET_OUT_LINUX ?= $(shell pwd)/out/linux_amd64
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-export TARGET_OUT = /work/out/$(TARGET_OS)_$(TARGET_ARCH)
-export TARGET_OUT_LINUX = /work/out/linux_amd64
+export TARGET_OUT = /work/src/istio.io/istio/out/$(TARGET_OS)_$(TARGET_ARCH)
+export TARGET_OUT_LINUX = /work/src/istio.io/istio/out/linux_amd64
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
 IMG ?= gcr.io/istio-testing/build-tools:master-2019-12-15T16-17-48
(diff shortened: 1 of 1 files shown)


seflerZ:
@sdake Would you please take a look?

sdake:
@seflerZ - I have been on PTO for a few weeks - then had flu for a few weeks. Apologies for the long lag in the review. I will take a look right now.

Cheers
-steve

howardjohn:
One potential benefit of this is you can do things like reference other repos (for example, in go.mod `replace istio.io/api => ../api`). However, it might be slow to mount everything (or it might not - i have no clue)

files/Makefile
```diff
@@ -108,15 +111,17 @@ RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):$(GID) --rm \
 	-e TARGET_OUT="$(TARGET_OUT)" \
 	-e TARGET_OUT_LINUX="$(TARGET_OUT_LINUX)" \
 	-e USER="${USER}" \
+	-e T="${T}" \
+	-e GOBIN="/work/src/istio.io/istio/bin" \
```
sdake: This is not obvious. The environment variable GOBIN is explicitly set in the [Dockerfile environment](https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile#L400).  We do not desire to override it in the Makefile.

howardjohn: Also this needs to apply to all repos

files/Makefile
```diff
@@ -28,6 +28,9 @@
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
+# This is the "go test" options. Default empty value.
+export T ?=
```
sdake: happy to permit go test overrides.

files/Makefile
```diff
@@ -56,8 +59,8 @@ export TARGET_OUT ?= $(shell pwd)/out/$(TARGET_OS)_$(TARGET_ARCH)
 export TARGET_OUT_LINUX ?= $(shell pwd)/out/linux_amd64
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
-export TARGET_OUT = /work/out/$(TARGET_OS)_$(TARGET_ARCH)
-export TARGET_OUT_LINUX = /work/out/linux_amd64
+export TARGET_OUT = /work/src/istio.io/istio/out/$(TARGET_OS)_$(TARGET_ARCH)
```
sdake: using src/istio.io/istio doesn't feel like the correct solution. The correct solution would take into account the environment in the test cases and expect binaries or test files in the correct location.

files/Makefile
```diff
@@ -108,15 +111,17 @@ RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):$(GID) --rm \
...
 	-v /etc/passwd:/etc/passwd:ro \
 	$(DOCKER_SOCKET_MOUNT) \
 	$(CONTAINER_OPTIONS) \
-	--mount type=bind,source="$(PWD)",destination="/work" \
+	--mount type=bind,source="$(PWD)",destination="/work/src/istio.io/istio" \
 	--mount type=volume,source=go,destination="/go" \
 	--mount type=volume,source=gocache,destination="/gocache" \
 	$(CONDITIONAL_HOST_MOUNTS) \
-	-w /work $(IMG)
+	-w /work/src/istio.io/istio $(IMG)
```
sdake: my knee-jerk reaction to this line of change as well as the other src/istio.io/istio changes is they are not the correct approach. I do not disagree that current testing is busted. I think the correct solution to solving the testing problems within a build container involve reading the target environment variable (or as it may be transposed in the istio/istio/Makefile.core.mk file... to a different variable) - to determine the correct testing behavior rather than adding hardcodes. I will give it some more thought. I am avialble on slack for further conversation.

files/Makefile
howardjohn: FWIW https://github.com/istio/common-files/pull/172 allows arbitrary env

sdake:
> One potential benefit of this is you can do things like reference other repos (for example, in go.mod `replace istio.io/api => ../api`). However, it might be slow to mount everything (or it might not - i have no clue)

bind mounting is cheap resource-wise.

files/Makefile
seflerZ: Well, it is because some legacy unit tests within the Pilot tried to resolve the source root path by matching "/src/istio.io/istio". If we don't intend to do modification here, we have to change the unit tests within Pilot.

This may be a little hard for we don't have any specified pattern to match. For example, if we rest in "[arbitrary path]/pilot/src/core/[here]", how can we tell the root source path since we have replaced "istio" directory name to "work"? If we choice to math "work", what if someone else mapped the "istio" to "mywork" like "mywork/pilot/src/..."?

files/Makefile
seflerZ: I fully understand. But the reality is even you choose to build with your local tool chain, you can't pass unit tests if your source code don't resident in "src/istio.io/istio"(Last time I made a soft link to "src/istio.io/istio" with "istio" and the tests failed). So we all agreed that it is the test code's responsibility to figure out the right path no matter where they have been mapped, but we need time to do refactoring. The problem is, if we don't make changes, we can't event make further testing and deployment.

files/Makefile
seflerZ: The $GOBIN was set to default "/gobin" value in the docker environment because the "Makefile" itself tried to read the $GOBIN but the docker env was isolated to the host. Therefore, what we've set in the Makefile won't take effect. Since we've produced Envoy binary in "work/bin", the test cases can not find the binary in "/gobin". We must explicitly pass in the $GOBIN env when using docker.

seflerZ: As you can see in Makefile line 133~134:
```
$(info Building with your local toolchain.)
export GOBIN ?= $(GOPATH)/bin
include Makefile.core.mk
```
When you're building local build chain, you use the "$GOPATH/bin" as the "$GOBIN", but in docker building path, we don't set any value and uses the default "/gobin". As far as I know, we produce the Envoy binary in "$TARGET_OUT/bin" mainly for debug purpose, if we put that in the path of docker image like "/gobin", we can't reach it after finishing building.

seflerZ: If this file was intent to be applied to all projects of Istio projects(like Galley, Mixer or something else). It's not wise to pass in "src/istio.io/istio" explicitly. We can take some similar approach, such as "src/istio.io/$project_name".

seflerZ: I know the default value for GOBIN is "/gobin". The problem is, in "Makefile.core.mk" the ISTIO_BIN was set to GOBIN which is different from GOPATH(which is "/go" in default value). However, we don't produce Envoy binary in "/gobin" but in "/go". This will cause the following problem:
```
--- FAIL: TestEnvoy (0.56s)
    xds_test.go:116: Failed to setup test: fork/exec /gobin/envoy: no such file or directory
```

ericvn: I don't think I like this as it may cause changes in the local _bin_ directory. And which _envoy_ would be copied into _bin_, the local OS (MacOS in my case) or the linux version since the container is using linux? I believe there is another issue which I put a change in the PR comments #19937 for where the linux envoy is polluting the darwin directory structure when building with the container, and this could pollute higher.

sdake: @ericvn in all cases where `BUILD_WITH_CONTAINER=1`, GOBIN should contain Linux binaries.

files/Makefile
sdake: @seflerz we are headed towards making `BUILD_WITH_CONTAINER=1` the default within istio/istio. Anything that detracts from that objective IMO is not ideal. The fact that the test cases need refactoring don't have much if any bearing - other than the work needs to be done, even if it needs to be forced.

files/Makefile
seflerZ: @howardjohn This is "go test" options, not env. Have you planed to cover it in #172 ? If it is, I'll close this PR.

howardjohn: Its an environment variable that we read and pass to `go test`, so it should be covered in #172

seflerZ: @howardjohn I'll close this PR.

PR #178, opened 2020-01-19
Add a BLOB target out for internet downloaded files

sdake:


sdake:
Depends-On: https://github.com/istio/tools/pull/683

PR #181, opened 2020-01-27
Update proxy builder tag

bianpengyuan:


bianpengyuan:
Yeah, but I'd rather keep the image tag in makefile override and makefile consistent. And do you know where is the common file update automator written (https://github.com/istio/proxy/pull/2635) It won't work for proxy because it does not update the proxy builder image. We might need to add some special logic to the automator for proxy.

howardjohn:
./tools/automator/automator.sh in test-infra

PR #189, opened 2020-02-04
Add a `make tidy-docker` make target

sdake:
As we revise the build-tools image, old images are cached and left
behind. `make tidy-docker` prunes docker stuff that we may not want
to persist when developer systems start to experience disk pressure.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -113,6 +113,6 @@ check-clean-repo:
 	@common/scripts/check_clean_repo.sh
 
 tidy-docker:
-	@docker images | grep gcr.io/istio-testing/build-tools | while read c1 c2 c3; do docker rmi -f $c1:$c2; done
+	@docker images | grep gcr.io/istio-testing/build-tools | while read c1 c2 c3; do docker rmi -f $$c1:$$c2; done
 
 .PHONY: lint-dockerfiles lint-scripts lint-yaml lint-copyright-banner lint-go lint-python lint-helm lint-markdown lint-sass lint-typescript lint-protos lint-all format-go format-python format-protos update-common update-common-protos lint-licenses dump-licenses dump-licenses-csv check-clean-repo tidy-docker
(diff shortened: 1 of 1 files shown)


sdake:
This has not been tested with BUILD_WITH_CONTAINER=1. I will test and remove the hold if things work as I expect they should.

files/common/Makefile.common.mk
```diff
@@ -112,4 +112,7 @@ update-common-protos:
 check-clean-repo:
 	@common/scripts/check_clean_repo.sh
 
-.PHONY: lint-dockerfiles lint-scripts lint-yaml lint-copyright-banner lint-go lint-python lint-helm lint-markdown lint-sass lint-typescript lint-protos lint-all format-go format-python format-protos update-common update-common-protos lint-licenses dump-licenses dump-licenses-csv check-clean-repo
+tidy-docker:
+	@docker images | grep gcr.io/istio-testing/build-tools | while read c1 c2 c3; do docker rmi -f $c1:$c2; done
```
howardjohn: I don't care much but if you prefer we do this in a different way elsewhere:
`docker images "gcr.io/istio-testing/build-tools/*" --format '{{.Repository}}:{{.Tag}}' | xargs -n1 docker rmi -f`

ericvn: I do it similarly as `docker rmi -f $(docker images --filter=reference="gcr.io/istio-testing/build-tools" -q)`

sdake: Two cool techniques. Wasn't aware of `--filter`. The `xargs` is a little difficult to understand.

IDC much how we do it fwiw, just that it is done. 👍

howardjohn: I think I have xargs Stockholm syndrome.

sdake: @ericvn - not clear that the technique you propose will work in a Makefile without a bash subshell. Will investigate.

@howardjohn - can you link to the code that uses that syntax above? I agree we should have consistency in the code base.

howardjohn: https://github.com/istio/istio/blob/2d44ff4f0225e228d64525e72f9c90fe3bd2aa11/prow/lib.sh#L81

PR #197, opened 2020-02-20
Add a troubleshooting link

sdake:
The build system has specific system requirements. As this technology
is new to Istio 1.5, some developers may not be familiar with the
requirements or how to ensure the requirements are met. Point all
developers to a troubleshooting page to distribute the developement
of the build system.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -68,6 +68,7 @@ UID = $(shell id -u)
 GID = `grep '^docker:' /etc/group | cut -f3 -d:`
 PWD = $(shell pwd)
 
+$(info If you suffer an unexpected failure, please reference: https://github.com/istio/istio/wiki/Troubleshooting-Development-Environment)
 $(info Building with the build container: $(IMG).)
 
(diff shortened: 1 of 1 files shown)


files/Makefile
```diff
@@ -68,6 +68,7 @@ UID = $(shell id -u)
 GID = `grep '^docker:' /etc/group | cut -f3 -d:`
 PWD = $(shell pwd)
 
+$(info If you suffer an unexpected failure, please reference: https://github.com/istio/istio/wiki/Troubleshooting-Development-Environment)
```
howardjohn: nit: is there any makefile magic to output a message when there is an error?

sdake: I gave this some consideration.

The main problem folks run into is an OOM situation. This OOM kills a bunch of random stuff, so any magic we might use may sometimes be lost. I'd also like to be a little more overt about the new nature of the development system. I'd rather have a bunch of bugs filed that are known with no solution than a bunch of bugs that are unknown.

PR #200, opened 2020-03-05
Fix for yaml-lint

nmittler:
yaml lint breaks on istio.io when -o pipefail is set. This is due to the fact that grep fails when it find no files containing templates "{{". This is a workaround to fix the istio.io build.


ericvn:
Note that after this change, I am seeing 
```
/bin/bash: @find: command not found
```
messages in the lint-yaml output.

PR #203, opened 2020-03-07
Fix build when using non container build

howardjohn:
I broke this in #172 because I forgot about the BUILD_WITH_CONTAINER=0
case... This should get things moving again. I have tested this in
https://github.com/istio/istio/pull/21951


howardjohn:
original PR was reverted, will try again

PR #207, opened 2020-03-09
Attempt 2 at docker run script

howardjohn:
WIP until we can verify on osx

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -30,103 +30,13 @@ SHELL := /bin/bash
 # figure out all the tools you need in your environment to make that work.
 export BUILD_WITH_CONTAINER ?= 0
 
-# Name of build container image
-IMAGE_NAME ?= build-tools
-
-# Version of image used within build container
-IMAGE_VERSION ?= master-2020-03-05T18-27-04
-
-LOCAL_ARCH := $(shell uname -m)
-ifeq ($(LOCAL_ARCH),x86_64)
-    TARGET_ARCH ?= amd64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
-    TARGET_ARCH ?= arm64
-else ifeq ($(LOCAL_ARCH),aarch64)
-    TARGET_ARCH ?= arm64
-else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 4),armv)
-    TARGET_ARCH ?= arm
-else
-    $(error This system's architecture $(LOCAL_ARCH) isn't supported)
-endif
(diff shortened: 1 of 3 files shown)


howardjohn:
To try out this PR, pull down https://github.com/istio/istio/pull/21997

^ also tests it passes in CI

ericvn:
Noted TMPDIR was needed on MacOS and @howardjohn fixed that in the prior commit. 

LGTM

files/Makefile
```diff
@@ -136,11 +44,20 @@ MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
...
 
 else
 
-$(info Building with your local toolchain.)
+# If we are not in build container, we need a workaround to get environment properly set
+# Write to file, then include
+$(shell mkdir -p out)
+$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
+include out/.env
+export out/.env
```
clarketm: Is this valid syntax? 
```diff
- export out/.env
+ export
```

files/Makefile
```diff
@@ -136,11 +44,20 @@ MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
...
+	@$(RUN) /bin/bash
+
 .PHONY: default
 
 else
 
-$(info Building with your local toolchain.)
+# If we are not in build container, we need a workaround to get environment properly set
+# Write to file, then include
+$(shell mkdir -p out)
```
clarketm: This fails `check-clean-repo` target. Can we change this to `mktemp` or add `out/` to all .gitignore files?

howardjohn: All repos are putting output in `out/` so we should add it to .gitignore everywhere. I sent it to a couple repos but must have missed some

files/Makefile
howardjohn: Good catch, I think this worked when I looked since istio/istio redundantly sets all of these. Testing locally what you suggested makes this work, Ill send a PR in a minute (testing it in https://github.com/istio/istio/pull/22018/files#diff-b67911656ef5d18c4ae36cb6741b7965)

sdake: I've commented this odd difference between bash and Makefile so people don't change it around without understanding the implications. See: https://github.com/istio/common-files/pull/211

files/Makefile
clarketm: An alternative is to use a _common_ .gitignore; something like: https://github.com/istio/common-files/pull/210 could simplify to distribution across the various **istio/*** repos

PR #208, opened 2020-03-09
PWD is not always set in CI

howardjohn:


files/Makefile
```diff
@@ -32,7 +32,7 @@ export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
 
-$(shell $(PWD)/common/scripts/setup_env.sh)
+$(shell $(shell pwd)/common/scripts/setup_env.sh)
```
clarketm: If it is not set in CI, then should this be fixed on line 57 where `BUILD_WITH_CONTAINER=0`?

PR #209, opened 2020-03-10
Fix make for CI

clarketm:
see: https://github.com/istio/common-files/pull/207#discussion_r390026663 for d415621.
see: https://github.com/istio/common-files/pull/208 for 4533a0d

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -54,9 +54,9 @@ else
 # If we are not in build container, we need a workaround to get environment properly set
 # Write to file, then include
 $(shell mkdir -p out)
-$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
+$(shell $(shell pwd)/common/scripts/setup_env.sh envfile > out/.env)
 include out/.env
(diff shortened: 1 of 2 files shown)


howardjohn:
feel free to remove the hold

files/Makefile
```diff
@@ -54,9 +54,9 @@ else
 # If we are not in build container, we need a workaround to get environment properly set
 # Write to file, then include
 $(shell mkdir -p out)
-$(shell $(PWD)/common/scripts/setup_env.sh envfile > out/.env)
+$(shell $(shell pwd)/common/scripts/setup_env.sh envfile > out/.env)
 include out/.env
-export out/.env
+export
```
sdake: is this stray debug?

clarketm: It is to export the included variables from the previous line. I think this was the original intention of `export out/.env`.

clarketm: Thanks for documenting this behavior in: https://github.com/istio/common-files/pull/211

PR #210, opened 2020-03-10
common .gitignore file

clarketm:
Support ignoring specific files/dirs when using `make` (e.g. `make check-clean-repo`):

* https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_gogo-genproto/46/gencheck_gogo-genproto/25
* https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_pkg/122/gencheck_pkg/60
* ...

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -51,6 +51,9 @@ else
     exit 1
 fi
 
+# Setup common .gitignore
+git config --local core.excludesfile ./common/config/.gitignore_common
+
 # Build image to use
 if [[ "${IMAGE_VERSION:-}" == "" ]]; then
   export IMAGE_VERSION=master-2020-03-05T18-27-04
--- a/files/common/config/.gitignore_common
+++ b/files/common/config/.gitignore_common
@@ -1,0 +1,2 @@
+# Contains built artifacts
+out/


clarketm:
> is this in addition to normal gitignore?

Yes, this it is additive.

files/common/scripts/setup_env.sh
```diff
@@ -51,6 +51,9 @@ else
     exit 1
 fi
 
+# Setup common .gitignore
+git config --local core.excludesfile ./common/config/.gitignore_common
```
howardjohn: The concern I see with this is it will modify the local environment when run outside of the container. I personally already have this setting set, so this would be mangling my setup.

Additionally, I suspect most people are using git outside of the container (direct git commands), so this won't apply there and they may get weird.

If we have common/.gitignore does it only apply to files inside of `common/`, or could we hack that to have 2 "real" gitignores?

clarketm: > The concern I see with this is it will modify the local environment when run outside of the container. I personally already have this setting set, so this would be mangling my setup.

Yes, the fact that only one file can exist for `core.excludesfile` at each level would indeed be mangling the setup if the user already has it set.

> If we have common/.gitignore does it only apply to files inside of common/, or could we hack that to have 2 "real" gitignores?

One option is to merge (i.e append) the `common/.gitignore` to the local `.gitignore` on `make update common`. Alternatively, the best I can think of is the alias `git` to `git -c ./common/.gitignore` inside `Makefile`'s.

clarketm:
I like the idea, but the execution is flawed. I will reopen once I have a more robust solution.

PR #214, opened 2020-03-11
remove yaml-lint hack

clarketm:
fix: https://github.com/istio/istio/issues/21854

Adding the `|| true` to prevents `grep` from emitting exit status **1** when no lines were selected _and_ the subsequent `xargs` call from emitting **123** b/c the previous command (i.e `grep`) emitted a exit status between **1-125**. 

> see `man bash`, `man grep` and `man xargs` for details.

**Before:**
```console
$ BUILD_WITH_CONTAINER=1 make lint-yaml ; echo $?
/bin/bash: @find: command not found
0
```
**After:**
```console
$ BUILD_WITH_CONTAINER=1 make lint-yaml ; echo $?
...
./pkg/kube/inject/testdata/inject/deploymentconfig.yaml
./pkg/kube/inject/testdata/inject/enable-core-dump.yaml
0
```

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -28,9 +28,8 @@ lint-dockerfiles:
 lint-scripts:
 	@${FINDFILES} -name '*.sh' -print0 | ${XARGS} shellcheck
 
-# TODO(nmittler): disabled pipefail due to grep failing when no files contain "{{". Need to investigate options.
 lint-yaml:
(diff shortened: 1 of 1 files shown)


howardjohn:
My understand was `-q` doesn't change the exit code:

```
$ grep -q notfound go.mod; echo $?
1
```

Am I missing something?

clarketm:
> My understand was `-q` doesn't change the exit code:
> 
> ```
> $ grep -q notfound go.mod; echo $?
> 1
> ```
> 
> Am I missing something?

Maybe I missed something in the local test case. LEet me look more into it.
/hold

PR #222, opened 2020-03-25
Fix recursive make

howardjohn:
Currently any targets calling make are broken because it tries to do
docker multiple layers in. This makes it so once we are in the docker
container, we don't try to go into another docker container.

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -96,6 +96,9 @@ if [[ -d "${HOME}/.kube" ]]; then
   CONDITIONAL_HOST_MOUNTS+="--mount type=bind,source=${HOME}/.kube,destination=/home/.kube "
 fi
 
+# Avoid recursive calls to make from attempting to start an additional container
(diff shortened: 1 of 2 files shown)


files/common/scripts/setup_env.sh
```diff
@@ -96,6 +96,9 @@ if [[ -d "${HOME}/.kube" ]]; then
   CONDITIONAL_HOST_MOUNTS+="--mount type=bind,source=${HOME}/.kube,destination=/home/.kube "
 fi
 
+# Avoid recursive calls to make from attempting to start an additional container
+export BUILD_WITH_CONTAINER=0
```
sdake: @howardjohn - I had a solution like this originally. There was a good reason for the IN_BUILD_CONTAINER environment variable, although I was unable to reproduce the problem @geeknoid had pointed out as he has removed the original work before I used a different variable (IN_BUILD_CONTAINER).

See PR: https://github.com/istio/common-files/pull/91#issuecomment-537496450 for more context.

PR #225, opened 2020-04-08
Update license linter

howardjohn:
Required if we merge https://github.com/istio/tools/pull/867

This adds a couple whitelists (for false negatives - they are acceptable licenses) and a few new license types that the new tool identifies

### DIFF
--- a/files/common/config/license-lint.yml
+++ b/files/common/config/license-lint.yml
@@ -9,6 +9,7 @@ unrestricted_licenses:
   - BSD-1-Clause
   - BSD-2-Clause
   - BSD-3-Clause
+  - 0BSD
   - FTL
   - LPL-1.02
   - MS-PL
@@ -23,6 +24,7 @@ unrestricted_licenses:
 
 reciprocal_licenses:
   - CC0-1.0
+  - CC-BY-3.0
   - APSL-2.0
   - CDDL-1.0
   - CDDL-1.1
@@ -32,6 +34,7 @@ reciprocal_licenses:
   - MPL-1.0
   - MPL-1.1
   - MPL-2.0
+  - MPL-2.0-no-copyleft-exception
   - Ruby
 
 restricted_licenses:
@@ -146,6 +149,16 @@ whitelisted_modules:
 
   # The core library is MIT licensed, but a submodule imports a BSD-3 license that licensee fails to correctly identify
   - github.com/vektah/gqlparser
+
+  # Apache 2.0
+  - github.com/spf13/cobra
+  - github.com/spf13/afero
+
+  # BSD-3 by pointer to another repo (github.com/gonum/license/blob/master/LICENSE)
+  - gonum.org/v1/plot
+
+  # FTL (https://github.com/golang/freetype/blob/master/licenses/ftl.txt)
+  - github.com/golang/freetype
 
   - github.com/ziutek/mymysql
   - gopkg.in/check.v1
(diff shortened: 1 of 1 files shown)


files/common/config/license-lint.yml
```diff
@@ -23,6 +24,7 @@ unrestricted_licenses:
 
 reciprocal_licenses:
   - CC0-1.0
+  - CC-BY-3.0
```
ericvn, line 27: Is there someone who verifies these licenses? Or when do we verify these additional licenses are OK?

howardjohn: I am not sure how we *should* do this, but I have been looking at googles internal license page for what is allowed/not allowed

ericvn: I was looking at IBM's well, and see some issues with CC-BY-x licenses which is why I picked on this particular one (CC-BY-3.0 wasn't specifically listed).

howardjohn: CC-BY-3.0 is explicitly listed as OK for us, while other CC-... are not

files/common/config/license-lint.yml
```diff
@@ -147,6 +151,23 @@ whitelisted_modules:
...
+  - github.com/spf13/afero
+
+  # BSD-3 by pointer to another repo (github.com/gonum/license/blob/master/LICENSE)
+  - gonum.org/v1/plot
+
+  # FTL (https://github.com/golang/freetype/blob/master/licenses/ftl.txt)
+  - github.com/golang/freetype
+
+  # CC-BY-3
+  - github.com/ajstarks/svgo
```
ericvn, line 165: I do see at the latest, `Update license to CC-BY-4.0`

howardjohn: The reason these had to be added is because the old license linter has a false negative -- the new one in my PR correctly identifies this as CC-BY-3 which is correct. We do no need CC-By-4

ericvn:
I'm fine with the changes, but I'm not sure what the Istio process if approving changes and if adding licenses should be reviewed by someone else earlier in the dev cycle or if it just gets reviewed and scanned at release time. For the CC-BY-3.0 and it's use, I believe that's do to the new license checker (so it's build time vs something we ship) so should be fine.

howardjohn:
The licenses are supposed to be automatically reviewed by the license linter. however, some repos like to do weird things which make the license finder not be able to identify the license. For these cases we need to manually whitelist.

ericvn:
Correct, but I assume the linter is checking the licenses it finds against how they are defined in the file. If we inadvertently mark a license unrestricted when they really need to be verified is my concern. From the discussion, those licenses added in the unrestricted area are at least approved by Google or used by Kubernetes within their open source.

howardjohn:
To clarify "unrestricted" means that the license allows us to do whatever we want essentially, compared to the ones in reciprocal which we can use but need to take some precautions. Internally we have a similar mapping of different categories which this is based off of. So when you say "when they really need to be verified is my concern" - the verification here is mostly copying from that internal doc. I am open to better suggestions though, but thats the extent of my legal knowledge :slightly_smiling_face:

ericvn:
@howardjohn I understand placing the new licenses in the appropriate category based on your internal doc. I was hoping that someone else would be able to identify if that was the appropriate thing for Istio (as opposed to checking against another source). From what I can tell looking at our internal lists, nothing here contradicts that. Just trying to make sure we don't include something that would make someone not want to use Istio based on licensing.

PR #232, opened 2020-04-21
Add check-no-modify target to prevent some files from being modified

ostromart:
assets.gen.go should not be updated in the repo after it is locally generated. This target supports testing this in a precommit check.


files/common/scripts/nomodify.md5
```diff
@@ -0,0 +1 @@
+cf4ad6158a6c85157286f3f17bbc7e9e  ../../operator/pkg/vfs/assets.gen.go
```
howardjohn, line 1: Why do we have repo specific stuff in common files, rather than in istio/istio?

PR #234, opened 2020-04-22
build-tools: use to latest master tag

clarketm:


ericvn:
I was going to add a comment that it's nice to know what the updates in the newer images are without having to go over to tools and figure out what's been committed between the times.

clarketm:
> I was going to add a comment that it's nice to know what the updates in the newer images are without having to go over to tools and figure out what's been committed between the times.

I agree. But doing this manually is tedious. If we can automate this then it would make this task a lot easier by appending a diff in the description. https://github.com/istio/test-infra/issues/2469

PR #237, opened 2020-04-28
Set up the conditional mounts correctly

sdake:
In prior implementations, there has always been a confusion between
what exactly $HOME is.
There are three variants of $HOME experienced in the field:
1. $HOME set to /home
1. $HOME set to /home/sdake
1. $HOME set to /Users/sdake

Not all of these conditions were properly managed. This PR handles the
three conditions above.

Related: https://github.com/istio/istio/pull/23293/

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -90,16 +90,25 @@ do
   unset -f "${f}"
 done
 
-# Set up conditional host mounts for docker and kubernetes config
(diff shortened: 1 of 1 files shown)


ericvn:
Note that this will break people who have a KUBECONFIG that contains multiple kube config files.

ericvn:
$HOME should always point to the users home directory. I'm not sure how this PR does anything to manage this.

PR #248, opened 2020-05-11
Update release 1.6 build tools

jasonwzm:


jacob-delgado:
@jasonwzm Do you believe this will be fixed https://github.com/istio/proxy/pull/2840?

jasonwzm:
@jacob-delgado No... I think Travis comment mentioned that the build failure is caused by an issue known for a while.

PR #254, opened 2020-05-30
All of our dockerfiles ignore DL3008

sdake:
I do not consider DL3008 to be a best practice. While it would be nice
to pin every single package version introduced in the distro, this is
not practical as old versions are removed from the repositories. If
we pin all versions we use, we risk a situation in which we can no
longer build.

The community appears to concur, as DL3008 is escaped in every dockerfile
where a version pin could be used.

### DIFF
--- a/files/common/config/.hadolint.yml
+++ b/files/common/config/.hadolint.yml
@@ -6,6 +6,7 @@
 # "make update-common".
 
 ignored:
+  - DL3008
 
 trustedRegistries:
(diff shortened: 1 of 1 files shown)


ericvn:
Will this change sounds correct to me, I know there was a lot of discussion in the past where a specific version pins was desired. One such after a quick look was https://github.com/istio/istio/pull/17239.

PR #256, opened 2020-05-30
Follow K8s loading pattern in makefiles

sdake:
```
from `kubectl config`:

 The loading order follows these rules:

  1.  If the --kubeconfig flag is set, then only that file is loaded. The flag may only be set once
and no merging takes place.
  2.  If $KUBECONFIG environment variable is set, then it is used as a list of paths (normal path
delimiting rules for your system). These paths are merged. When a value is modified, it is modified
in the file that defines the stanza. When a value is created, it is created in the first file that
exists. If no files in the chain exist, then it creates the last file in the list.
  3.  Otherwise, ${HOME}/.kube/config is used and no merging takes place.
```

This PR handles spaces in directories and duplicate-like naming of config files.

This code works by creating a random number and using that random number as the mount
for the kubeconfig.

Co-Authored-By: Eric Van Norman <ericvn@us.ibm.com>

Big thanks to Eric for the non-stop testing of this work on various platforms and finding
all of the previous flaws in the implementation.

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -92,6 +92,7 @@ done
 
 # Set conditional host mounts
 export CONDITIONAL_HOST_MOUNTS=${CONDITIONAL_HOST_MOUNTS:-}
+container_kubeconfig=''
 
 # docker conditional host mount (needed for make docker push)
 if [[ -d "${HOME}/.docker" ]]; then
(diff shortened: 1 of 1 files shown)


ericvn:
I probably shouldn't approve as a co-author but I did run my non KUBECONFIG, KUBECONFIG set to one value and KUBECONFIG set to 3 configs tests on this and they all passed.

files/common/scripts/setup_env.sh
```diff
@@ -103,12 +104,45 @@ if [[ -d "${HOME}/.config/gcloud" ]]; then
...
+# testcase: "a b:c d:e f"
+# testcase: "a b:c:d e"
+parse_KUBECONFIG () {
+TMPDIR=""
+if [[ "$1" =~ ([^:]*):(.*) ]]; then
+  while true; do
+    rematch=${BASH_REMATCH[1]}
+    kubeconfig_random="$(od -vAn -N4 -tx /dev/random | tr -d '[:space:]' | cut -c1-8)"
+    container_kubeconfig+="/home/${kubeconfig_random}:"
+    CONDITIONAL_HOST_MOUNTS+="--mount type=bind,source=${rematch},destination=/config/${kubeconfig_random},readonly,consistency=delegated "
```
brian-avery, line 124: Looking at the consistency flag (https://docs.docker.com/storage/bind-mounts/), consistency delegated says:



delegated: The container runtime’s view of the mount is authoritative. There may be delays before updates made in a container are visible on the host.

cached: The macOS host’s view of the mount is authoritative. There may be delays before updates made on the host are visible within a container.

Since this is mounted readonly, should this be cached?

brian-avery: Spoke with Steve off PR. See this PR for why delegated was used: https://github.com/istio/common-files/pull/252. TL;DR: performance increases

PR #260, opened 2020-06-05
Add fix-copyright-banner

howardjohn:
This will alleviate some pain when creating new files

### DIFF
--- a/files/common/scripts/fix_copyright_banner.sh
+++ b/files/common/scripts/fix_copyright_banner.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# WARNING: DO NOT EDIT, THIS FILE IS PROBABLY A COPY
+#
+# The original version of this file is located in the https://github.com/istio/common-files repo.
+# If you're looking at this file in a different repo and want to make a change, please go to the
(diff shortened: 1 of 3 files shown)


files/common/scripts/fix_copyright_banner.sh
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+WD=$(dirname "$0")
+WD=$(cd "$WD"; pwd)
+
+for fn in "$@"; do
+  if ! grep -L -q -e "Apache License, Version 2" -e "Copyright" "${fn}"; then
```
mandarjog, line 30: Check for Istio authors not just Apache 2.

howardjohn: If I do this, the lint check fails since some files do *not* have "Istio Authors". It may be a legitimate issue (I am not familiar with copyright laws, so no clue) but its not changed by my PR. This check is identical to the lint script, and should be kept in sync there. I think its outside the scope of this PR to change the linter

sdake: PR LGTM

PR #261, opened 2020-06-19
Only mount/add kubeconfig in build container if it exists

ericvn:
The reason for this change was we actually need more that Docker and make to use the build container. We actually needed a kubeconfig file (could be fake) that could be mounted in the build container. None of the developer docs talked about how to create this container and some developers, say docs, might not need to actually have a kube cluster configured. This change should allow us to go back to not requiring a kubeconfig file being available when running the build container.

I tested this by renaming my ~/.kube/config file so it wouldn't be found, and well as adding a fictitious kubeconfig file in a set of them (admittedly only at the end).

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -104,7 +104,7 @@ if [[ -d "${HOME}/.config/gcloud" ]]; then
   CONDITIONAL_HOST_MOUNTS+="--mount type=bind,source=${HOME}/.config/gcloud,destination=/config/.config/gcloud,readonly,consistency=delegated "
 fi
 
-# Thus function checks if the file exists. If it does, it creates a randomly named host location
+# This function checks if the file exists. If it does, it creates a randomly named host location
 # for the file, adds it to the host KUBECONFIG, and creates a mount for it.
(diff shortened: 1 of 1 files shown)


ericvn:
@sdake @rcaballeromx Here is a PR that should verify if a config exists before trying to create the mount point for it and adding to KUBECONFIG in the build container.

rcaballeromx:
@ericvn Thank you for making this change.

files/common/scripts/setup_env.sh
```diff
@@ -104,6 +104,16 @@ if [[ -d "${HOME}/.config/gcloud" ]]; then
   CONDITIONAL_HOST_MOUNTS+="--mount type=bind,source=${HOME}/.config/gcloud,destination=/config/.config/gcloud,readonly,consistency=delegated "
 fi
 
+# Thus function checks if the file exists. If it does, it creates a randomly named host location
```
brian-avery, line 107:
```suggestion
# This function checks if the file exists. If it does, it creates a randomly named host location
```

brian-avery:
This looks good to me. I'll let @sdake give the final approval.

ericvn:
@sdake Any thoughts?

PR #265, opened 2020-06-29
Update .golangci.yml

stevenctl:
Updating the config file to tell golangci-lint to use the latest version 
see: https://golangci-lint.run/usage/configuration/#config-file

1.27 is already installed in the latest build-tools
```
docker run gcr.io/istio-testing/build-tools:master-2020-06-25T05-18-39 golangci-lint version
golangci-lint has version v1.27.0 built from (unknown, mod sum: "h1:VYLx63qb+XJsHdZ27PMS2w5JZacN0XG8ffUwe7yQomo=") on (unknown)
```

Keeping `do-not-merge` while testing that this won't break `make go-lint` in istio/istio

### DIFF
--- a/files/common/config/.golangci.yml
+++ b/files/common/config/.golangci.yml
@@ -7,7 +7,7 @@
 
 service:
   # When updating this, also update the version stored in docker/build-tools/Dockerfile in the istio/tools repo.
-  golangci-lint-version: 1.21.x # use the fixed version to not introduce new linters unexpectedly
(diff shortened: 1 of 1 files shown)


howardjohn:
Even if it breaks istio the commit won't merge and then we can fix it there so not a big deal. Early testing is always good to, but not as bad as the changes tlto the build container which are NOT tested before merhe

stevenctl:
Works fine but doesn't fix what I'm seeing in https://github.com/istio/istio/pull/25046

PR #271, opened 2020-07-08
Better handle vendored trees

jwendell:
- Skip the vendor dir in `find` usages
- Create a target for running `go mod download` to be reused by other
  projects, that work on vendored trees.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -19,7 +19,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses \) -prune -o -type f
+FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses -o -path ./vendor \) -prune -o -type f
 XARGS = xargs -0 -r
 
 lint-dockerfiles:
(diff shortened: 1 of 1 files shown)


jwendell:
Part of https://github.com/istio/istio/issues/25330

files/common/Makefile.common.mk
```diff
@@ -64,15 +64,17 @@ lint-typescript:
...
+lint-licenses: mod-download-go
 	@license-lint --config common/config/license-lint.yml
 
 lint-all: lint-dockerfiles lint-scripts lint-yaml lint-helm lint-copyright-banner lint-go lint-python lint-markdown lint-sass lint-typescript lint-protos lint-licenses
 
 tidy-go:
 	@go mod tidy
 
+mod-download-go:
+	@-GOFLAGS="-mod=readonly" go mod download
```
howardjohn, line 76: For my understanding, can you explain why we need mod=readonly? My understanding is that makes the go.mod file not modified here. Maybe that is useful, but I don't see how it relates to vendor directory. Does it somehow?

jwendell: in go 1.14, if there's a vendor dir, `-mod=vendor` is implied unless explicitly overwritten.
also, users may have the `GOFLAGS` globally set with `-mod=vendor`.

So, overwriting this with `readonly` on the command line seems to be the safest thing to do, that works on both scenarios without any side effect (like updating go.mod).

jwendell: also, tools like `go list -m all` do not work in vendor mode. After this gets merged/propagated I'll submit a PR for istio fixing it.

howardjohn: Oh so this more about unsettting GOFLAGs than it is about setting it to readonly?

jwendell: Yes, the choices are `readonly` or `mod`. I find readonly more conservative, less intrusive.

PR #273, opened 2020-07-09
Change caching to `cached` in support of new file synchronization

ericvn:
This leaves file caching enabled in a recent PR but turns off two-way syncing on the Mac: https://docs.docker.com/docker-for-mac/mutagen/#bypassing-a-two-way-sync-for-a-volume


ericvn:
@sdake Any thoughts on this? Should we turn back the dial part way, or do we revert back to the default consistency?

PR #275, opened 2020-07-10
set common make shell to bash

tbarrella:
Previously, `make lint-yaml` in the tools repo would fail with
> /bin/sh: 1: set: Illegal option -o pipefail

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -19,6 +19,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


howardjohn:
@nmittler I recall you spent quite a bit of time on this issue. Is this a sufficient fix?

ericvn:
Fixes the issue when using `zsh` on a Mac. Only concern is we are forcing `/bin/bash` to exist on a developers machine. I'm not sure if we have developers that use Windows and what happens there. Most of the Linux related OS's are probably fine.

PR #284, opened 2020-07-21
Use /home/.cache as a volume mount

sdake:
The proxy build uses /home/.cache to store envoy proxy output. Use that
directory as a volume mount to store the output of the build results.
This takes a 20 minute build to 12 seconds on my gear. This does not
solve the problem that the proxy build does not output binaries,
however, it gets Istio closer to that goal.

### DIFF
--- a/files/common/scripts/run.sh
+++ b/files/common/scripts/run.sh
@@ -50,6 +50,6 @@ export REPO_ROOT=/work
     --mount "type=bind,source=${PWD},destination=/work,consistency=cached" \
     --mount "type=volume,source=go,destination=/go,consistency=cached" \
     --mount "type=volume,source=gocache,destination=/gocache,consistency=cached" \
-    --mount "type=volume,source=cache,destination=/home/.cache,consistency=delegated" \
+    --mount "type=volume,source=cache,destination=/home/.cache,consistency=cached" \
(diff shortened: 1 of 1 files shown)


sdake:
Part of: https://github.com/istio/istio/issues/25572

sdake:
may be an error, let me double check. Thanks for the quick review.

files/common/scripts/run.sh
```diff
@@ -50,5 +50,6 @@ export REPO_ROOT=/work
     --mount "type=bind,source=${PWD},destination=/work,consistency=cached" \
     --mount "type=volume,source=go,destination=/go,consistency=cached" \
     --mount "type=volume,source=gocache,destination=/gocache,consistency=cached" \
+    --mount "type=volume,source=cache,destination=/home/.cache,consistency=delegated" \
```
ericvn, line 53: Do we want this to be `delegated` based on the upcoming changes with two-way sync and Mutagen? 

Maybe it makes sense to then do cache and gocache as delegated as well? I know it took a while to set up the source directory with two way for the initial pass (although there may be improvements in later releases).

sdake: delegated, cached, no consistency all work properly on Linux. On Mac, YMMV - one platform at a time.

PR #288, opened 2020-07-23
New image version

sdake:


ericvn:
Verifying this actually fixes the problem.

PR #300, opened 2020-08-15
Use the variable VERSION, if it is defined

jwendell:
The variable ISTIO_VERSION isn't defined anywhere (actually it is in release-builder
repo, but it's not needed, I'll remove it) but VERSION is defined in Makefile.core.mk.

This allows developer builds to print something like this:
```
$ ./out/linux_amd64/istioctl version --short=false --remote=false
version.BuildInfo{Version:"1.8-dev", GitRevision:"96f07bf388573cb94d5d92a26c058ed321bc4447",...
```
Instead of duplicating the git revision.

Users are able to override this at build time with
```
$ VERSION=1.8-my-feature make build
```

### DIFF
--- a/files/common/scripts/report_build_info.sh
+++ b/files/common/scripts/report_build_info.sh
@@ -35,17 +35,11 @@ if [[ -z "${IGNORE_DIRTY_TREE}" ]] && ! git diff-index --quiet HEAD --; then
   tree_status="Modified"
 fi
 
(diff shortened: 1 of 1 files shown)


jwendell:
Should be merged after https://github.com/istio/istio/pull/26543

jwendell:
related pr: https://github.com/istio/release-builder/pull/379

PR #303, opened 2020-08-24
WIP: Fix get_protos so that it actually pulls updated protos rather than f…

sdake:
get_protos is not pulling all protos during a make. Fix that.

### DIFF
--- a/get_protos.sh
+++ b/get_protos.sh
@@ -78,22 +78,18 @@ popd >/dev/null || exit
 popd >/dev/null || exit
 
 # Retrieve a copy of K8s api proto files
-echo "k8s.io/api"
+echo "kubernetes/api"
 pushd "${TEMPDIR_API}" >/dev/null || exit
 git clone -q --single-branch --branch master https://github.com/kubernetes/api.git
-pushd api >/dev/null || exit
 git checkout -q "${API_TAG}"
-popd >/dev/null || exit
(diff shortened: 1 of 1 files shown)


get_protos.sh
```diff
@@ -78,22 +78,18 @@ popd >/dev/null || exit
 popd >/dev/null || exit
 
 # Retrieve a copy of K8s api proto files
-echo "k8s.io/api"
+echo "kubernetes/api"
 pushd "${TEMPDIR_API}" >/dev/null || exit
 git clone -q --single-branch --branch master https://github.com/kubernetes/api.git
-pushd api >/dev/null || exit
```
brian-avery, line 84: Does this work? We do a git clone and then try to checkout the tag without cding into the repo.

get_protos.sh
```diff
@@ -78,22 +78,18 @@ popd >/dev/null || exit
...
 git checkout -q "${API_TAG}"
-popd >/dev/null || exit
 find . -name \*proto | cpio --quiet -pdm "${REPODIR}"/common-protos/k8s.io
 popd >/dev/null || exit
 
 # Retrieve a copy of K8s apimachinery proto files
-echo "k8s.io/apimachinery"
```
brian-avery, line 91: Does this work? We do a git clone and then try to checkout the tag without cding into the repo.

sdake:
Zombie PR. Closing.

PR #304, opened 2020-08-31
Add bash helper function to find the right way to do inplace sed commands

ssuchter:
Because sed -i has different arguments on Mac (BSD) versus Linux (GNU), this function figures out the right invocation and puts it into an environment variable.

See istio/istio.io#8033 for the motivating reason and how I will use it in another script.

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -163,3 +163,31 @@ if [[ "${1}" == "envfile" ]]; then
   echo "TARGET_ARCH=${TARGET_ARCH}"
   echo "BUILD_WITH_CONTAINER=0"
 fi
+
+# TODO: Use $INPLACE_SED in every script that uses the -i option on sed
+find_inplace_sed() {
+    local tmpfile
+    local POSSIBLE_INPLACE_SED_CMDS
+    local sed_cmd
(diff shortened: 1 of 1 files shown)


ssuchter:
@ericvn @sdake @jwendell  I think this is the right place now, PTAL.

ericvn:
I'm not trying to minimize the work put into this, but we do have the build container set up so that developers do not need to install a lot of special software, etc on their laptops. The idea was that `make` and `docker` should allow those with Linux and Mac machines to do development in an easy manner. For the majority of the Istio repos (proxy might be the only exception but not sure), one simply runs `make xyz` which starts the build container, invokes make with the specified target and runs the various scripts and executables within the container. Output is directed to the host for binaries, images, etc. Sometimes new targets may be needed.

This work seems to go against using the build container, and may require the developer to install additional software locally; things like linkinator, htmlproofer, and hugo. And the developer should use the same versions of these files as the pipeline, which if they are working on multiple projects may or may not be the case.

All that said, I'm fine adding this function, I'm just not sure we want to complicate scripts more needed.

ericvn:
I see istio/istio.io#8803 was closed so closing this one as well. @ssuchter Feel free to reopen if these should be considered again.

PR #307, opened 2020-09-04
Create passwd entry and group entry on startup of container

sdake:
This doesn't add much overhead, but ensures credentials are
consistent across platforms, specially MacOS and Linux.

This happens by not mounting passwd nor groups, and instead adding
these files at startup of the docker container.

Depends-On: https://github.com/istio/tools/pull/1196


mandarjog (approved):
Lgtm
Nit: the commit msg and pr msg is not accurate.
It should be
“Do not Mount passwd and group files from host”

PR #309, opened 2020-09-10
Trim and document license allowlist

howardjohn:
A lot of this stuff is not needed anymore

Blocked by https://github.com/istio/istio.io/pull/8118

Verified in istio/istio, istio/test-infra, and istio/tools


howardjohn:
This is ready to go now. I verified in tools, test-infra, istio, and istio.io this will pass. Down from 104 to 14 items in allowlist.

brian-avery (approved):

PR #310, opened 2020-09-10
Deduplicate kind cluster setup and refactor cluster topology configuration into a file to define arbitrary topologies

su225:
Currently script to setup kind clusters are almost the same in istio/istio and istio/istio.io repositories. This PR aims to unify and remove duplication. We also introduce the concept of cluster configuration file with which we can unify both single and multicluster setup and define topology in a flexible manner. 

Here is an example for creating 3 kind clusters.
```json
[
  {
      "cluster_name": "cluster1",
      "pod_subnet": "10.10.0.0/16",
      "svc_subnet": "10.255.10.0/24",
      "network_id": 0
  },
  {
      "cluster_name": "cluster2",
      "pod_subnet": "10.20.0.0/16",
      "svc_subnet": "10.255.20.0/24",
      "network_id": 0
  },
  {
     "cluster_name": "cluster3",
      "pod_subnet": "10.30.0.0/16",
      "svc_subnet": "10.255.30.0/24",
      "network_id": 1
  }
]
```
**NOTE**: Validation of cluster topology configuration is not performed and this is for testing purposes only.

### DIFF
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -51,13 +51,35 @@ function load_cluster_topology() {
     exit 1
   fi
 
-  readarray -t CLUSTER_NAMES < <(jq -r '.[].cluster_name' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
-  readarray -t CLUSTER_POD_SUBNETS < <(jq -r '.[].pod_subnet' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
-  readarray -t CLUSTER_SVC_SUBNETS < <(jq -r '.[].svc_subnet' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
-  readarray -t CLUSTER_NETWORK_ID < <(jq -r '.[].network_id' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
-  
+  export CLUSTER_NAMES
+  export CLUSTER_POD_SUBNETS
+  export CLUSTER_SVC_SUBNETS
+  export CLUSTER_NETWORK_ID
+
+  while read -r value; do
+    CLUSTER_NAMES+=("$value")
+  done < <(jq -r '.[].cluster_name' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
+
+  while read -r value; do
+    CLUSTER_POD_SUBNETS+=("$value")
+  done < <(jq -r '.[].pod_subnet' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
+
+  while read -r value; do
+    CLUSTER_SVC_SUBNETS+=("$value")
+  done < <(jq -r '.[].svc_subnet' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
+
+  while read -r value; do
+    CLUSTER_NETWORK_ID+=("$value")
+  done < <(jq -r '.[].network_id' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
+
   export NUM_CLUSTERS
   NUM_CLUSTERS=$(jq 'length' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
+
+  echo "${CLUSTER_NAMES[@]}"
+  echo "${CLUSTER_POD_SUBNETS[@]}"
+  echo "${CLUSTER_SVC_SUBNETS[@]}"
+  echo "${CLUSTER_NETWORK_ID[@]}"
+  echo "${NUM_CLUSTERS}"
 }
 
 #####################################################################
@@ -77,6 +99,8 @@ function cleanup_kind_cluster() {
 }
 
 # setup_kind_cluster creates new KinD cluster with given name, image and configuration
+# Even in case of single cluster, don't call this directly. Call through setup_kind_clusters
+# with cluster topology configuration file.
 # 1. CLUSTER_CONFIG: KinD cluster configuration YAML file (mandatory)
 # 2. NAME: Name of the Kind cluster (optional)
 # 3. IMAGE: Node image used by KinD (optional)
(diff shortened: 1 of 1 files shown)


su225:
Here is one way how `kind_provisioner.sh` can be called
```bash
set -e
set -x

source "files/common/scripts/kind_provisioner.sh"

ARTIFACTS="test_artifacts"
IMAGE="kindest/node:v1.18.2"
CLUSTER_YAML="trustworthy-jwt.yaml"
IP_FAMILY="ipv4"
METRICS_SERVER_CONFIG_DIR="./test_metrics/metrics"


if [[ ${MULTICLUSTER} = 1 ]]; then
  TOPOLOGY_FILE="cluster_config.json"
else
  TOPOLOGY_FILE="cluster_config_single.json"
fi

if [[ ! -f "${TOPOLOGY_FILE}" ]]; then
  echo 'topology file not found..exiting'
  exit 1
fi

load_cluster_topology "${TOPOLOGY_FILE}"
setup_kind_clusters "${IMAGE}" "${CLUSTER_YAML}" "${IP_FAMILY}"
```

files/common/scripts/kind_provisioner.sh
```diff
@@ -0,0 +1,319 @@
...
+#  { "cluster_name": "cluster3","pod_subnet": "10.30.0.0/16","svc_subnet": "10.255.30.0/24","network_id": "1" }]
+function load_cluster_topology() {
+  CLUSTER_TOPOLOGY_CONFIG_FILE="${1}"
+
+  if [[ ! -f "${CLUSTER_TOPOLOGY_CONFIG_FILE}" ]]; then
+    echo 'cluster topology configuration file is not specified'
+    exit 1
+  fi
+
+  readarray -t CLUSTER_NAMES < <(jq -r '.[].cluster_name' "${CLUSTER_TOPOLOGY_CONFIG_FILE}")
```
nmittler, line 54: have you verified this works? Will we need to add `jq` to the docker image?

su225: There are istioio tests which use `jq` already. Here are two examples I found
https://github.com/istio/istio.io/blob/master/content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
https://github.com/istio/istio.io/blob/master/content/en/docs/tasks/observability/metrics/tcp-metrics/test.sh

su225: Any issue with `readarray`? I use bash 4.4.20. Here is output for `bash --version`
```bash
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
```

stevenctl: I think something like the following is preferred. Pretty sure OSX ships with bash 3, and some users running locally may be on zsh. 
```
while read value; do
    CLUSTER_NAMES +=($value)
done < $CLUSTER_TOPOLOGY_CONFIG_FILE
```

files/common/scripts/kind_provisioner.sh
```diff
@@ -0,0 +1,319 @@
...
+  declare -a KUBECONFIGS
+  for CLUSTER_NAME in "${CLUSTER_NAMES[@]}"; do
+    KUBECONFIG_FILE="${KUBECONFIG_DIR}/${CLUSTER_NAME}"
+    if [[ ${NUM_CLUSTERS} -gt 1 ]]; then
+      install_metallb "${KUBECONFIG_FILE}"
+    fi
+    KUBECONFIGS+=("${KUBECONFIG_FILE}")
+  done
+
+  ITER_END=$((NUM_CLUSTERS-1))
```
stevenctl, line 210: nice!

stevenctl (commented):
Looks good. I wish we had a way to confirm this runs right via CI, and doesn't break tests.  

I guess that's one issue with moving this to common-files: it will be harder to see if changes to this script will break CI. Is it possible to add a Prow job for this repo to see that least deploys KinD without breaking? 

To make sure we don't break actual tests though.. we'd have to copy the changes we want to make to the repo, make a dummy PR to trigger Prow, iterate there and move the finally copy here. A pretty hacky workaround I guess..

Merging this initial version shouldn't cause any harm since nothing is pointing to it.

files/common/scripts/kind_provisioner.sh
```diff
@@ -0,0 +1,319 @@
...
+
+# setup_kind_cluster creates new KinD cluster with given name, image and configuration
+# 1. CLUSTER_CONFIG: KinD cluster configuration YAML file (mandatory)
+# 2. NAME: Name of the Kind cluster (optional)
+# 3. IMAGE: Node image used by KinD (optional)
+# 4. IP_FAMILY: valid values are ipv4 and ipv6.
+# This function returns 0 when everything goes well, or 1 otherwise
+# If Kind cluster was already created then it would be cleaned up in case of errors
+function setup_kind_cluster() {
+  CLUSTER_CONFIG_YAML=${1:-"./prow/config/trustworthy-jwt.yaml"}
```
stevenctl, line 87: If we don't move these configs to common-files we should have them be environment variables that are required to be set at the time of `source kind_provisioner.sh`.

files/common/scripts/kind_provisioner.sh
nmittler: I'm pretty sure our scripts already require at least bash 4.x. I seem to have updated mine to 5.x.

files/common/scripts/kind_provisioner.sh
su225: :)

su225:
> To make sure we don't break actual tests though.. we'd have to copy the changes we want to make to the repo, make a dummy PR to trigger Prow, iterate there and move the finally copy here. A pretty hacky workaround I guess..

Yeah. This is fine. Once this is merged here, we can make that dummy PR a real one, test it again and merge it.

files/common/scripts/kind_provisioner.sh
su225: Or at least at the time of calling `setup_kind_clusters`?

files/common/scripts/kind_provisioner.sh
su225: Anyways for maximum compatibility, I'm changing it to the way @stevenctl  suggested.

files/common/scripts/kind_provisioner.sh
su225: As this is meant to be used as a library, I am adding check to make sure that `DEFAULT_CLUSTER_YAML` is defined before invoking this.

nmittler (approved):

PR #311, opened 2020-09-11
KUBECONFIGS should be exported so that it can be used downstream

su225:
We will use in `prow/integ-suite-kind.sh` to create a `:` separated string `KUBECONFIG` to feed it to `setup_env.sh` script so that the required kubeconfig files are mounted to the container properly.

### DIFF
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -225,7 +225,7 @@ EOF
 
   # Install MetalLB for LoadBalancer support. Must be done synchronously since METALLB_IPS is shared.
   # and keep track of the list of Kubeconfig files that will be exported later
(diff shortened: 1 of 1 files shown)


su225:
@nmittler - I had missed this one.

su225:
Experimenting with `integ-suite-kind.sh` changes when mounted properly it looks like this
```
build-tools:/work$ ls -l /config
total 24
-rw------- 1 user 1000 5419 Sep 11 14:33 184a8de3
-rw------- 1 user 1000 5423 Sep 11 14:33 4d2a58d5
-rw------- 1 user 1000 5419 Sep 11 14:33 895fffbd
```
(by the way, this is real. I'll create a PR against istio.io repo shortly)

nmittler (approved):

PR #331, opened 2020-10-21
[release-1.8] Automated branching step 4

GregHanson:


howardjohn (approved):
are you sure this doesn't need `UPDATE_BRANCH`? won't it override UPDATE_BRANCH in other repos?

ericvn:
Looking @howardjohn, as I think it should

PR #333, opened 2020-10-26
Use build image with gen-release-notes

sdake:


ericvn (approved):

ericvn (changes requested):
Maybe you want the second build from today: master-2020-10-26T17-20-43

PR #351, opened 2020-11-16
automatically detect whether to run `docker run -it` or just `docker run`

yskopets:
Signed-off-by: Yaroslav Skopets <yaroslav@tetrate.io>

## Summary

* to be able to run Docker-based commands on GitHub Actions, automatically detect when it's possible to run `docker run -it` and when it should be just `docker run`

### DIFF
--- a/files/common/scripts/run.sh
+++ b/files/common/scripts/run.sh
@@ -37,11 +37,18 @@ export REPO_ROOT=/work
 MOUNT_SOURCE="${MOUNT_SOURCE:-${PWD}}"
 MOUNT_DEST="${MOUNT_DEST:-/work}"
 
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
I think this makes sense. @sdake @ericvn thoughts?

BTW - not sure how feasible this is but rather than using DinD you may consider just using the build tools image as your base image on github actions, then disabling BUILD_WITH_CONTAINER. That is what we do in our CI

yskopets:
@howardjohn yeah, I started by trying to avoid DinD as you described.

But in the end I still need Docker options from `run.sh` only in the form of GitHub Actions YAML.

So, it's just simpler to accept DinD than maintain lots of boilerplate in GitHub Actions YAML.

ericvn (approved):

PR #370, opened 2021-01-15
Update image for new docs protoc plugin

howardjohn:
This should not be merged until https://github.com/istio/api/pull/1821 is, as they are co-dependent


howardjohn:
API PR is approved by TOC, so whoever approves this should approve that and we can get this in

ericvn (approved):

PR #378, opened 2021-01-25
[release-1.8] improve cluster config file reader

stevenctl:
unblocks https://github.com/istio/istio/pull/30365


stevenctl:
added do-not-merge, need to make a test PR against istio/istio to make sure this is backwards compat

stevenctl:
Seems like this passes in release-1.8, I think those tests are pretty flaky. The PR this unblocks will reduce those flakes.

bianpengyuan (approved):

PR #387, opened 2021-02-24
enable core dumps in kind

howardjohn:
Goes with https://github.com/istio/istio/pull/31049. Should make sure
that passes tests before merge


mandarjog (approved):
Lgtm

mandarjog:
Holding since the linked pr fails unit test.

PR #406, opened 2021-04-20
Update base image for 1.10

brian-avery:
There were two builds on April 6. We got the first build but missed the second.

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -63,7 +63,7 @@ fi
 
 # Build image to use
 if [[ "${IMAGE_VERSION:-}" == "" ]]; then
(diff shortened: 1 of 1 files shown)


Monkeyanator (approved):

ericvn:
Pointer to an automation tweak I guess. This second update wasn't needed as the image didn't change. The reason a second image got created on 4/6 was do to updating the build image within the script to a new one. So, we will always be one image behind.

What I think happens:
- Someone updates the Dockerfile in istio/tools.
- Automation creates a new image (this is good)
- We manually check for the new image tag, and then update common-files with the new tag.
- Many repos get the new common-file update with the new image (in setup-env.sh)
- As part of the post-submit for tools, it sees the script is changed and automation kicks of a new image build. (this one isn't needed).

PR #420, opened 2021-05-20
Revert to kind v0.10 from v0.11

jacob-delgado:
See https://github.com/istio/tools/pull/1540


jacob-delgado:
See https://github.com/istio/common-files/pull/419

PR #424, opened 2021-05-25
Retry when something that should never happen does

howardjohn:


howardjohn:
merged in https://github.com/istio/common-files/pull/425/files

PR #428, opened 2021-05-27
Bump to kind v0.11.0

howardjohn:


files/common/scripts/setup_env.sh
```diff
@@ -63,7 +63,7 @@ fi
 
 # Build image to use
 if [[ "${IMAGE_VERSION:-}" == "" ]]; then
-  export IMAGE_VERSION=master-2021-05-27T17-28-48
+  export IMAGE_VERSION=master-2021-05-27T21-48-48
```
mandarjog, line 66: Can you point to build tools commit that has version 0.11.0?

howardjohn: https://github.com/istio/tools/pull/1553/files

mandarjog (approved):
Lgtm

ericvn:
Note that we are currently using kind 0.11.1 with the latest image.

ericvn (changes requested):
We have moved to a newer version of kind, so merging this will revert that change.

PR #432, opened 2021-06-08
files/Makefile: allow custom command with shell target

rmstoi:
This commit adds feature to Makefile to run other commands besides
/bin/bash in build-tools container. By default /bin/bash will run, but
that can be overridden with CMD variable.

Running custom commands besides bash is useful for automation of tasks
in scripts. For example, this command would return output path of bazel

BUILD_WITH_CONTAINER=1 make shell CMD='bazel info output_path'

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -39,6 +39,7 @@ export
 $(shell $(shell pwd)/common/scripts/setup_env.sh)
 
 RUN = ./common/scripts/run.sh
+CMD ?= /bin/bash
 
 MAKE_DOCKER = $(RUN) make --no-print-directory -e -f Makefile.core.mk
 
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Seems `common/scripts/run.sh <whatever you want>` may be sufficient?

rmstoi:
> Seems `common/scripts/run.sh <whatever you want>` may be sufficient?

For istio proxy repo run.sh by default would pull istio-testing/build-tools, whereas with make it would be istio-testing/build-tools-proxy. That is due to setting IMAGE_NAME in [Makefile.overrides.mk](https://github.com/istio/proxy/blob/master/Makefile.overrides.mk#L17) of istio proxy. Hence, correct way to start the build-tools docker image for proxy would be:

`IMAGE_NAME=build-tools-proxy common/scripts/run.sh`

IMHO, handling this via makefile simplifies things and removes the need to explicitly copy overrides from Makefile.overrides.mk to the command line.

PR #436, opened 2021-06-21
Add integfuzz tag to lint and format configs

kailun-qin:
Partially addresses https://github.com/istio/istio/issues/33418

Signed-off-by: Kailun Qin <kailun.qin@intel.com>


ericvn:
Will need the istio/istio change in https://github.com/istio/istio/pull/33541 merged before merging this to all the automated update to propagate without causing failures. Has this change been tried in the other repos?

howardjohn (approved):

PR #438, opened 2021-06-23
Remove unsupported golangci-lint setting for gofumpt simplify

kailun-qin:
The setting to simplify code is not supported by `golangci-lint` according to
https://github.com/golangci/golangci-lint/blob/505ed3c0f7fc4941a76e52c00e462ad38ecb79ec/.golangci.example.yml#L261.

This was for `gofmt` previously.

Signed-off-by: Kailun Qin <kailun.qin@intel.com>

### DIFF
--- a/files/common/config/.golangci.yml
+++ b/files/common/config/.golangci.yml
@@ -72,9 +72,6 @@ linters-settings:
   golint:
     # minimal confidence for issues, default is 0.8
     min-confidence: 0.0
-  gofumpt:
-    # simplify code: gofmt with `-s` option, true by default
-    simplify: true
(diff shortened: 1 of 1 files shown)


kailun-qin:
Note that even it's going to be supported in the future release of `gofumpt`: https://github.com/mvdan/gofumpt/issues/115, it should align with the current `gofumpt`'s behavior which enforces `gofmt`'s simplify flag (defaults to true where we don't need to set it explicitly).

kailun-qin:
PTAL, thanks! @howardjohn

howardjohn (approved):

PR #464, opened 2021-08-19
Add script and make target to update IMAGE_VERSION to most recent image

ericvn:
This `make` target can then be used by automation to update the setup_env.sh after a new build-tools image is created.

Part of https://github.com/istio/test-infra/issues/2469

### DIFF
--- a/bin/update_build_image.sh
+++ b/bin/update_build_image.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -e
+
+# Get the current release of the build-tools image by looking at IMAGE_VESION in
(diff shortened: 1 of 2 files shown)


bin/update_build_image.sh
```diff
@@ -0,0 +1,33 @@
...
+# ../files/common/scripts/setup_env.sh
+imageRelease=$(grep IMAGE_VERSION= files/common/scripts/setup_env.sh | sed -e 's/^.*=//' | cut -f1,2 -d'-')
+
+# If this isn't a release- branch, cut after the first -
+if [[ "$imageRelease" != "release-"* ]]; then
+  imageRelease=$(echo "$imageRelease" | cut -f1 -d'-')
+fi
+
+# Get the latest build-tools image for the given release
+newBuildImage=$(curl -sL https://gcr.io/v2/istio-testing/build-tools/tags/list | jq '."manifest"[]["tag"]' | awk '/'$imageRelease'/ && !/latest/' | sort -r | sed  -e 's/^[[:space:]]*"//' -e 's/".*//' | head -n 1)
```
howardjohn, line 28: does it return everything? no paging, etc?

howardjohn: if this is complex it could take it as an arg and istio/tools can pass it as an arg (maybe)?

ericvn: The curl returns a lot of tags (over 1400) so I don't see any paging. I originally had this coded up as a way to simply run the `make` locally, and then manually create a PR with the latest image.

I certainly can extend this script to use a passed in parameter/env var for the new IMAGE_VERSION. If it's set, use that value instead of the curl.

ericvn: @howardjohn Some changes to use an arg/env var to pass in the value. If nothing is passed in, use the original logic.

PR #485, opened 2021-11-03
fix kind create cluster failed on linux when do integrate test manually

zackzhangkai:
Synced from https://github.com/istio/istio/pull/35862/files.

Signed-off-by: zackzhangkai <zhangkaiamm@gmail.com>

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -72,7 +72,7 @@ tidy-go:
 	@go mod tidy
 
 mod-download-go:
-	@-GOFLAGS="-mod=readonly" go mod download
+	@GOFLAGS="-mod=readonly" go mod download
 # go mod tidy is needed with Golang 1.16+ as go mod download affects go.sum
 # https://github.com/golang/go/issues/43994
 	@go mod tidy
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -37,6 +37,8 @@ DEFAULT_KIND_IMAGE="gcr.io/istio-testing/kindest/node:v1.19.1"
 # COMMON_SCRIPTS contains the directory this file is in.
 COMMON_SCRIPTS=$(dirname "${BASH_SOURCE:-$0}")
 
+KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
+
 # load_cluster_topology function reads cluster configuration topology file and
 # sets up environment variables used by other functions. So this should be called
 # before anything else.


files/common/Makefile.common.mk
```diff
@@ -72,7 +72,7 @@ tidy-go:
 	@go mod tidy
 
 mod-download-go:
-	@-GOFLAGS="-mod=readonly" go mod download
+	@GOFLAGS="-mod=readonly" go mod download
```
howardjohn, line 75: can you revert or explain why its changed?

files/common/scripts/kind_provisioner.sh
```diff
@@ -37,6 +37,8 @@ DEFAULT_KIND_IMAGE="gcr.io/istio-testing/kindest/node:v1.19.1"
 # COMMON_SCRIPTS contains the directory this file is in.
 COMMON_SCRIPTS=$(dirname "${BASH_SOURCE:-$0}")
 
+KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
```
howardjohn, line 40: Why is this needed? Isn't kind creating a new kubeconfig anyways?

zackzhangkai: Kind will create files on `/config/xxx`, which will have no permission. The errors is :

ERROR: failed to create cluster: failed to write KUBECONFIG: open /config/fb379ac4: read-only file system

files/common/Makefile.common.mk
zackzhangkai: It just wants to set the GOFLAGS variable, no need `-`.

More details , can run `make gen`  to get.

howardjohn: it does actually: https://stackoverflow.com/questions/28305609/what-does-prefix-mean-in-makefile

zackzhangkai: In the case '@-', if the command got failed, it will continue. But it will occur error on the going. The errors such as "should use GOFLAGS=readonly" or "go mod vendor". 

So maybe no '-' is more suitable.

ericvn: Looking back, the `-` seemed to be on `make lint-license` and not the other targets. In combining the target commands, it seems the `-` was adding to most of the targets. I'm not sure if that was intentional or not.

You might want to verify at least `make license-lint` works on other repos (I know it did on istio/istio when the change was there).

I agree that removing the `-` so the error isn't ignored seems to be the correct thing, unless we know a particular target expects the error and we need the `-` to ignore it.

files/common/scripts/kind_provisioner.sh
ericvn: I would need to see if this breaks what some are doing when running locally. Need to see if this pollutes the `config` file every time the command is run (ie. new configurations are added). When they get created in the config directory, it is easy to delete them after shutting down the clusters.

zackzhangkai: I can reproduce every time. 

do `make shell` to run a virtual environment(docker), then run `./prow/integ-suite-kind.sh` to start Kind clusters.

ericvn: @zackzhangkai That's exactly what we do. In our case we run something like ` ARTIFACTS=$PWD/artifacts ./prow/integ-suite-kind.sh --topology MULTICLUSTER --skip-build --skip-cleanup --topology-config prow/config/topology/external-istiod.json`. When complete, I can do a:
```
ls artifacts/
config-config.yaml  config-external.yaml  kind	kubeconfig  topology-config.json  trace.yaml
```
which has the kubeconfig.
I do make a copy of the topology file locally and enter the kubeconfig in that copy and use that on the go tests invocation.

PR #490, opened 2021-11-11
Remove useless shell script in makefile

SpecialYang:
Remove useless shell script `set_up_env.sh` in makefile. When we use make command with mode of `BUILD_WITH_CONTAINER=1`, the shell script set_up_env.sh will be executed twice. The first execution of `set_up_env.sh` in makefile does no effect and is unnecessary.

See https://github.com/istio/istio/pull/36016.

### DIFF
--- a/files/Makefile
+++ b/files/Makefile
@@ -32,11 +32,9 @@ export BUILD_WITH_CONTAINER ?= 0
 
 ifeq ($(BUILD_WITH_CONTAINER),1)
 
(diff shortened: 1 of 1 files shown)


SpecialYang:
@ericvn PTAL.

ericvn (approved):
Setting the environment variables within the `shell` are run in a new shell and not reflected.  The run.sh sources the setup_env so the env vars are set within that shell.

PR #518, opened 2022-01-25
Output junit view for linter

howardjohn:
Requires golangci-lint 1.44


jacob-delgado (approved):
Lgtm. Remove DNM when linting will work. I approved part of your PR for that

PR #519, opened 2022-01-26
Remove strict on helm linter (so it won't fail on deprecation warnings)

ericvn:
Similar to https://github.com/istio/istio/pull/36996, there is another lint failure in a post-submit: https://prow.istio.io/view/gs/istio-prow/logs/lint_tools_postsubmit/1486403734660452352.

This is because of a change in the Helm linter as we upgraded versions.

This makes a similar change to the overall helm linting for the project. I'm not sure if this makes sense for the whole Istio project.

I do not see a way to not error on a subset of warnings.


howardjohn:
I fixed it in istio/tools. I think this one only is needed on istio/istio which has its own target

ericvn:
OK. I see https://github.com/istio/tools/pull/1839 fixes it for istio/tools which is where I noted the failure above. Will close this PR.

PR #535, opened 2022-02-24
Backport changes from istio/istio for kind retries

howardjohn:


jacob-delgado (approved):
see https://github.com/istio/istio/pull/37438

PR #537, opened 2022-02-28
We trust in quay.io registry

jwendell:


howardjohn:
Where do we plan to use it? I don't personally trust quay for reliability - it caused plenty of issues when we had critical images there. So would not be in favor of having any core images there, but for other uses could be ok

jwendell:
I see. Maistra does use quay (for instance to use centos stream images), so I thought it'd be ok to whitelist it.

howardjohn (approved):
whitelist for linter sgtm. Just would prefer. not to use this in our core images

PR #540, opened 2022-03-01
Add support for kind dual-stack

Kmoneal:
First steps in https://github.com/istio/istio/issues/37533.

* Configure cluster IP family to support dual-stack
* Configure coredns to support dual-stack

### DIFF
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -191,44 +191,6 @@ EOF
     kubectl apply -f "${METRICS_SERVER_CONFIG_DIR}"
   fi
 
-    # update coredns to use respond to A queries and AAAA queries and not
-  # only the type of the first Cluster IP
-  if [ "${IP_FAMILY}" == "dual" ]; then
-    cat <<EOF | kubectl apply -f -
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: system:coredns
-rules:
-- apiGroups:
-  - ""
-  resources:
-  - endpoints
-  - services
-  - pods
-  - namespaces
-  verbs:
-  - list
-  - watch
-- apiGroups:
-  - discovery.k8s.io
-  resources:
-  - endpointslices
-  verbs:
-  - list
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - nodes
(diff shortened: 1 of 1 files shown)


files/common/scripts/kind_provisioner.sh
```diff
@@ -191,6 +191,44 @@ EOF
...
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+EOF
+    kubectl patch deployment coredns -n kube-system -p '{"spec":{"template":{"spec":{"containers":[{"name":"coredns","readinessProbe":{"httpGet":{"path":"/health","port":8080}}}]}}}}'
+    kubectl patch deployment coredns -n kube-system -p '{"spec":{"template":{"spec":{"containers":[{"name":"coredns", "image":"coredns/coredns:1.8.7"}]}}}}'
+    kubectl rollout restart -n kube-system deployment coredns
```
howardjohn, line 229: nit: this restarts 3 times. We can definitely remove the last one, but could also merge the first one?

howardjohn (commented):
are the coredns changes upstream in some version?

jacob-delgado:
> are the coredns changes upstream in some version?

CoreDNS 1.8.1+ is what fixed dual-stack support. See https://coredns.io/2021/01/20/coredns-1.8.1-release/

kind uses the default version of CoreDNS that is shipped with kubeadm. kubeadm did not use a version of CoreDNS that support dual-stack services until 1.22.0. We had to support previous versions of K8s pre-1.22 that did not contain an updated CoreDNS changes. @Kmoneal please remove the patches for CoreDNS as in open source we are only planning on supporting 1.22+ for dual-stack.

files/common/scripts/kind_provisioner.sh
```diff
@@ -191,6 +191,44 @@ EOF
...
+  fi
+
```
jacob-delgado, line 231:
```suggestion
```

files/common/scripts/kind_provisioner.sh
Kmoneal: Removed coredns because it's likely not needed

Kmoneal:
@nmittler we would appreciate any feedback, if possible

jacob-delgado (approved):

PR #541, opened 2022-03-03
add istio_feature.proto for feature https://docs.google.com/document/…

zhlsunshine:
According to [proposal design doc](https://docs.google.com/document/d/1ZcTIdHYNnLpwgLJa8tWg0YwsyHAmc_4lXguLmMQvAj0/edit?resourcekey=0-6OWzJ5cnlh_oYl4gFMbOuA#heading=h.xw1gqgyqs5b), add proto file for Istio feature management.

### DIFF
--- a/common-protos/istio.io/extensions/istio_feature.proto
+++ b/common-protos/istio.io/extensions/istio_feature.proto
@@ -1,0 +1,48 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


common-protos/istio.io/extensions/istio_feature.proto
```diff
@@ -0,0 +1,48 @@
+// Copyright 2019 Istio Authors
```
jacob-delgado, line 1: The year no longer gets included.

jacob-delgado (commented):
I feel like there should be a job in `test-infra` to update this proto when `istio/api` is updated. I understand that PR is still in-flight. Should the generated proto also be included?

common-protos/istio.io/extensions/istio_feature.proto
zhlsunshine: okay, removed

zhlsunshine:
> I feel like there should be a job in `test-infra` to update this proto when `istio/api` is updated. I understand that PR is still in-flight. Should the generated proto also be included?

Yes, agree, `test-infra` should aware it if `istio/api` has any change.

PR #554, opened 2022-03-14
write gencheck patch to artifacts

stevenctl:
Sometimes gen(check) is slow for me and I just want to apply whatever was generated by the script.


howardjohn (approved):
nice

PR #564, opened 2022-03-24
Remove override of REPO_ROOT in build container

ericvn:
This is the first part of fixing https://github.com/istio/istio/issues/38103.  `docker-builder` looks at REPO_ROOT to get the location to copy files. When used with a local release-builder build, this causes the docker files to be read from the release-builder repo where they do not exist.


johnma14 (approved):

howardjohn:
Do we need to add the REPO_ROOT logic back to istio/istio then?

PR #572, opened 2022-04-01
Fix `make build` in istio/istio using wrong Golang cache directory.

ericvn:
`make build` in istio/istio results in:
```
go: could not create module cache: mkdir /Users: permission denied
go: could not create module cache: mkdir /Users: permission denied
go: could not create module cache: mkdir /Users: permission denied
go: could not create module cache: mkdir /Users: permission denied
```

The BLOCKLIST change yesterday now allows `GOPATH=..` to be passed to the build container which results in this error. Specifically add GOPATH= to the BLOCKLIST.


howardjohn (approved):

nmittler:
@ericvn I'm still seeing:

```
> make docker push
./tools/docker
go: cannot find GOROOT directory: /Users/nathanmittler/go/go
```

Looks like we need to add `GOROOT` to this as well. I tried locally and it works.

PR #573, opened 2022-04-01
Fix make build in istio/istio

nmittler:
This was partially addressed by https://github.com/istio/common-files/pull/572, but was missing `GOROOT`.


ericvn (approved):
I didn't have GOROOT in my environment. Hopefully, this is the only remaining GO* variable. Note that GOROOT wouldn't have been blocked prior to the original regression.

PR #575, opened 2022-04-08
replace golint with revive

zirain:
replace golint with revive

part of https://github.com/istio/istio/issues/38201

need merge https://github.com/istio/istio/pull/38287 and https://github.com/istio/release-builder/pull/1023 first.

### DIFF
--- a/files/common/config/.golangci.yml
+++ b/files/common/config/.golangci.yml
@@ -40,7 +40,7 @@ linters:
   - gocritic
   - gofumpt
   - goimports
-  - golint
+  - revive
   - gosimple
   - govet
   - ineffassign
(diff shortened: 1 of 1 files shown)


files/common/config/.golangci.yml
```diff
@@ -70,6 +70,67 @@ linters-settings:
   golint:
```
ericvn, line 70: I think we can remove the `golint` section.

files/common/config/.golangci.yml
```diff
@@ -70,6 +70,67 @@ linters-settings:
...
     # minimal confidence for issues, default is 0.8
     min-confidence: 0.0
+  revive:
```
ericvn, line 73: In my earlier testing I had just used the same two lines for `revive` but I do like this calling out better.

Also, can we move this to the correct place alphabetically.

ericvn:
Add DNM until the prerequisites PRs are merged.

files/common/config/.golangci.yml
zirain: will fix the place later, also wait for John's agreement on istio repo changes.

files/common/config/.golangci.yml
zirain: ok, will remove later.

zirain:
@ericvn I think it's ok to get this merged.

ericvn:
It looks like there are other lint issues in a few of the other repos as well. I'll work on some updates.

PR #595, opened 2022-05-11
Remove GCI

hzxuzhonghu:
now goimports well grouped the import package


howardjohn (commented):
Source on goimports? I don't think it does what gci does last I saw. We don't just want grouping we want 100% deterministic ordering and grouping)

hzxuzhonghu:
I tried goimports in a private repo, seems it works.

hzxuzhonghu:
My fault, goimports can not group well

PR #597, opened 2022-05-12
fix git describe --tags command from returning fatal log messages

GregHanson:


ericvn (approved):
This allows the variable to be a shortened commit if a tag is not available.

PR #638, opened 2022-07-22
Simplify linter/format command

howardjohn:
It already runs against all files (regardless of if we just pass one file)


ericvn:
Agree that we do run against all files, regardless of what's passed. Should we add `--skip-files` for the files we thought we would ignore?

howardjohn:
it's already in the config yaml

ericvn (approved):

PR #676, opened 2022-09-30
Updates noted while converting istio/community to use common-files

ericvn:
Updating the istio/community repo to use common-files in https://github.com/istio/community/pull/799.

There are two comments in that PR:
 - One related to the date in the LICENSE file.
 - The other relates to the CONTRIBUTING.md file in the repos. As this common-files version of CONTRIBUTING.md simply points to the istio/community, we need to update the `update=-common` target to only update CONTRIBUTING if we are replacing a version that points to the community file. In the case of istio/community, we don't want to update the file.

For the later task, I could not determine a way to figure out which repo are are updating, so electing to only update the CONTRIBUTING.md if it contains a pointer to the istio/community file.

### DIFF
--- a/files/LICENSE
+++ b/files/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2016-2020 Istio Authors
+   Copyright 2016-2022 Istio Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


ericvn:
Making this DNM until the istio/community PR is merged and the second change is truly needed.

howardjohn (approved):

ericvn:
Let this merge. istio/community has an update that should work with the pipeline automated update, and the community repo was added into the pipeline in test-infra.

PR #687, opened 2022-10-18
Add build arch to build info

hanxiaop:


files/common/scripts/report_build_info.sh
```diff
@@ -44,3 +44,4 @@ echo "istio.io/pkg/version.buildGitRevision=${BUILD_GIT_REVISION}"
 echo "istio.io/pkg/version.buildStatus=${tree_status}"
 echo "istio.io/pkg/version.buildTag=${GIT_DESCRIBE_TAG}"
 echo "istio.io/pkg/version.buildHub=${HUB}"
+echo "istio.io/pkg/version.buildArch=${TARGET_ARCH}"
```
howardjohn, line 47: TARGET_ARCH is not always used; gobuild.sh is actually using BUILD_GOOS

Also do we want to make it linux/amd64 or something? Or add the linux part somehwere?

hanxiaop: @howardjohn Thanks

howardjohn (approved):

PR #703, opened 2022-11-03
fix: excludes Ruby env vars

bartoszmajsak:
When using tools such as `rvm` host env variables contain e.g. `GEM_HOME` and `GEM_PATH`. These might differ with what is expected by the container. 

As a consequence running tests lead to errors such as

```
build-tools:/work$ fpm
/usr/lib/ruby/vendor_ruby/rubygems.rb:265:in `find_spec_for_exe': can't find gem fpm (>= 0.a) with executable fpm (Gem::GemNotFoundException)
	from /usr/lib/ruby/vendor_ruby/rubygems.rb:284:in `activate_bin_path'
	from /usr/local/bin/fpm:25:in `<main>'
```

This PR excludes `rvm`-related variables.


howardjohn (approved):
/ok-to-test

PR #710, opened 2022-11-09
Copy change from istio/istio/41872 to prevent override

ericvn:
Also cherry pick to 1.16 as the mentioned PR was as well.


howardjohn (approved):
thanks, I forgot about this

PR #719, opened 2022-11-16
chore: clean proto related targets

daixiang0:
After #614, no proto format anymore.

Also, `lint-protos` only is a target, no rules to run.

Signed-off-by: Loong <loong.dai@intel.com>

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -63,7 +63,7 @@ lint-typescript:
 lint-licenses:
 	@if test -d licenses; then license-lint --config common/config/license-lint.yml; fi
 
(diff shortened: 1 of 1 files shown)


daixiang0:
`clang-format` can format protos:
```
  --style=<string>               - Coding style, currently supports:
                                     LLVM, GNU, Google, Chromium, Microsoft, Mozilla, WebKit.
                                   Use -style=file to load style configuration from
                                   .clang-format file located in one of the parent
                                   directories of the source file (or current
                                   directory for stdin).
                                   Use -style=file:<format_file_path> to explicitly specifythe configuration file.
                                   Use -style="{key: value, ...}" to set specific
                                   parameters, e.g.:
                                     -style="{BasedOnStyle: llvm, IndentWidth: 8}"
```

If accepted, I can add it.

howardjohn:
Please don't use clang-formart for protos, we use `buf` for everything proto related

PR #742, opened 2023-01-03
Update to use kind Kubernetes image 1.26.0 by default

ericvn:


howardjohn (approved):
note: istio/istio doesn't use this, should set gcr.io/istio-testing/kind-node:v1.25.0

PR #753, opened 2023-01-18
Add allowlisted license for sha1cd

ericvn:


files/common/config/license-lint.yml
```diff
@@ -123,3 +123,5 @@ allowlisted_modules:
 # MIT: https://github.com/invopop/yaml/blob/v0.1.0/LICENSE
 - github.com/invopop/yaml
 
+# Apache 2.0: https://github.com/pjbgf/sha1cd/blob/main/LICENSE
```
howardjohn, line 126: I think they recently added it. Where does the dep come from? can we use a newer version with the license?

ericvn:
Moving to a later commit with the actual LICENSE file. It was added post the last release. So this change isn't needed.

PR #777, opened 2023-02-07
Diasble mdl linting complaint of ending with single newline

ericvn:
The new mdl linter is complaining about the inconsistency of having a single newline at the end of the markdown files.

Disable the check for now.

Testing in istio/istio.io


ericvn:
Updating the files in the individual repos.

PR #782, opened 2023-02-15
detect and use arm image

stevenctl:
When running `run.sh` or `make` stuff, it should automatically use the native arch build-tools image.


ericvn (dismissed):

howardjohn:
isn't it a multi-arch image?

PR #810, opened 2023-03-27
allow post-processing linter config

howardjohn:
goes with https://github.com/istio/istio/pull/44130


files/common/config/.golangci.yml
```diff
@@ -12,8 +12,8 @@ run:
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   deadline: 20m
   build-tags:
-  - integ
-  - integfuzz
+    - integ
```
howardjohn, line 15: these changes are just auto-formatting so when we format later it is a NOP

ericvn (approved):
I think this is better to allowing a repo to add specific code than try and do various things in the common-files scripts depending on certain files being available in the repo.

PR #811, opened 2023-03-30
Added `ORG` to allow users to change the where common files sha comes from

Kmoneal:
Allow users to specify which org images and common-files sha comes from.


files/common/Makefile.common.mk
```diff
@@ -93,9 +93,11 @@ mirror-licenses: mod-download-go
 TMP := $(shell mktemp -d -u)
 UPDATE_BRANCH ?= "master"
 
+ORG=${ORG:-istio}
+
 update-common:
 	@mkdir -p $(TMP)
-	@git clone -q --depth 1 --single-branch --branch $(UPDATE_BRANCH) https://github.com/istio/common-files $(TMP)/common-files
+	@git clone -q --depth 1 --single-branch --branch $(UPDATE_BRANCH) https://github.com/$(ORG)/common-files $(TMP)/common-files
```
ericvn, line 100: What's the likelihood that someone would have an ORG env var set that this would cause issues. I know I don't, but would someone else?

Would there be something better? I haven't looked to see if we have something elsewhere in our repository code.

Kmoneal: We do have GITHUB_ORG but I can see this as something not everyone wants to change so how about BUILD_TOOLS_ORG?

Kmoneal: ORG  is now BUILD_TOOLS_ORG

ericvn:
As a test, I pulled the Makefile and setup_env change to my local machine and tried `make shell` and that works. A `make update-common` worked as well.

PR #812, opened 2023-04-03
Add codespaces to common files

therealmitchconnors:
After piloting codespaces in ztunnel, we are ready to make them available in all istio repos.  Once this is merged to repos, I will enable build caching so that the image used in code spaces always has a recent build ready to go.


files/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+  OS="$(uname | tr '[:upper:]' '[:lower:]')" && \
+  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" && \
+  KREW="krew-${OS}_${ARCH}" && \
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" && \
+  tar zxvf "${KREW}.tar.gz" && \
+  ./"${KREW}" install krew
+
+ENV PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
+
+RUN kubectl krew install ns
```
howardjohn, line 84: I think this could just be `apt install kubectx` without krew, unless we want it for other things

files/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,91 @@
+# Remote docker environment for ztunnel development.
+#
+# Build:
+#   docker build -t ztunnel/remote-env:0.1 -f docker/remote-env/Dockerfile .
```
howardjohn, line 4: should probably not be ztunnel anymore?

files/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# Clear credentials:
+#   ssh-keygen -f "$HOME/.ssh/known_hosts" -R "[localhost]:2222"
+#
+# stop:
+#   docker stop ztunnel-dev
+#
+# ssh credentials (test user):
+#   ssh user@localhost -p2222 (enter `password` on the command-line)
+
+FROM gcr.io/istio-testing/build-tools:master-42499d67179f80bbbeb71bc0f793d1b483cf8937
```
ericvn, line 23: I'm not sure if you want to stick to to a fixed build-tools. I'm thinking we could possibly update the pipeline logic to update this at the same time that we update the image in common-files.

howardjohn: Its also probably reasonable to publish this directly from istio/tools ?

ericvn: I don't know enough about codespaces if one needs to provide a Dockerfile can it call out an image directly json.

files/.devcontainer/Dockerfile
therealmitchconnors: unfortunately, that results in `E: Unable to locate package kubectx`.  If it worked, it would be way easier like that...

files/.devcontainer/Dockerfile
therealmitchconnors: I'm not sure I understand the discussion on this thread.  We could use a raw gcr.io/istio-testing/build-tools image without modification, but it won't work for ztunnel, and it won't include docker out of docker, kind, and the other tools we're adding here.  As for how to automatically update the version of build-tools we use, I am open to advice...

therealmitchconnors: I can specify the image tag as an arg, but that arg has to be specified at build time, and dev containers don't support setting args.  Even if they did, it would be a static value from the json file.  I think the best bet is to edit whatever command updates the build-tools tag that make uses, and add updating the dockerfile too.

files/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+  KREW="krew-${OS}_${ARCH}" && \
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" && \
+  tar zxvf "${KREW}.tar.gz" && \
+  ./"${KREW}" install krew
+
+ENV PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
+
+RUN kubectl krew install ns
+
+ENV BUILD_WITH_CONTAINER=0
```
ericvn, line 40: Could this be set in the devcontainer.json if using a normal image?

files/.devcontainer/Dockerfile
ericvn: > I mean we can make a build-tools-codespaces variant as part of publishing build-tools
The variant answers my question above as if we create a variant or is the change minimal enough adding it to build-tools. Sounds like we should just add another variant in the tools docker file.

ericvn:
Here are some changes, https://github.com/istio/tools/pull/2426, which adds the few items to the current build-tools image.

files/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "istio build-tools",
+  "image": "ericvn/build-tools:master-latest-amd64", // TODO: @ericvn update to official image
```
ericvn, line 3:
```suggestion
  "image": "gcr.io/istio-testing/build-tools:master-8c6fd99f50a8acae83a260340848097c002d2e0b",
```

therealmitchconnors:
Added recommended change to copy .devcontainer files.

ericvn:
I expect that merging this PR will likely only update the .devcontainer file to the repos , but will update the makefile in the repos copy the .devcontainer file with the next update_common. We have another common-files PR in the queue which will hopefully do that.

ericvn (approved):
At least get the makefile update to the other repos. The .devcontainer file may come in a follow-on update.

Approving since we need to get do two updates anyway.

PR #816, opened 2023-04-05
Update to only cp .devcontainer and .gitattributes on a Mac and not .* (seems to pick up parent)

ericvn:
I notice that running `make update-common` locally on a Mac after the last change causes my repository to get messed up due to the fact that it seems to also want to copy `..` with the `.*`. Not sure why the build-tools container would behavior differently on a Mac, but it does. 

Some testing before the prior change showed that the .devcontainer directory wasn't getting copied on a Mac. It appears the pipeline does copy it without special .<name> callouts.

So, we could revert back to not calling out the .<name> files which seems to copy the .files in the pipeline but not on a Mac locally. If we want the Mac to behave like the pipeline, it seems we need to call out the two copies. 

I'm fine with not incorporating this change, but if an Istio dev is using a Mac (and not sure what other platforms), they may run into tan issue that there local repository is messed up if the happen to run `make update-common`.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -102,7 +102,7 @@ update-common:
 	@if [ "$(CONTRIB_OVERRIDE)" != "CONTRIBUTING.md" ]; then\
 		rm $(TMP)/common-files/files/CONTRIBUTING.md;\
 	fi
-	@cp -a $(TMP)/common-files/files/* $(TMP)/common-files/files/.* $(shell pwd)
+	@cp -a $(TMP)/common-files/files/* $(TMP)/common-files/files/.devcontainer $(shell pwd)
 	@rm -fr $(TMP)/common-files
 	@$(or $(COMMONFILES_POSTPROCESS), true)
 
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -102,7 +102,7 @@ update-common:
 	@if [ "$(CONTRIB_OVERRIDE)" != "CONTRIBUTING.md" ]; then\
 		rm $(TMP)/common-files/files/CONTRIBUTING.md;\
 	fi
-	@cp -a $(TMP)/common-files/files/* $(TMP)/common-files/files/.devcontainer $(shell pwd)
+	@cp -a $(TMP)/common-files/files/* $(TMP)/common-files/files/.devcontainer $(TMP)/common-files/files/.gitattributes $(shell pwd)
 	@rm -fr $(TMP)/common-files
 	@$(or $(COMMONFILES_POSTPROCESS), true)
 


ericvn:
@howardjohn Thoughts on this? My concern is if we merge and it picks up the gitattributes will it cause issues. We can watch the pipeline and like DNM any bad looking PRs as they are created (if they would even be created).

Interestingly, the old code wouldn't have copied the .devcontainers but it is created in the repos now with the last PR so maybe the .gitattribites issue only happens locally, and the change might just work normally in the pipeline.

ericvn:
> Copying .gitattributes seems intended/desired? We can a common .gitattributes file

So maybe the old code just works without specifying the additional . files, although not what I was seeing locally. Or maybe we just never update that file across repos. Need to look into that.

howardjohn:
If I run `update-common` from istio master it looks fine - no changes made. What are you seeing?

ericvn:
So difference:
```
drwxr-xr-x@  3 ericvn  staff    96B Apr  5 12:33 .devcontainer
-rw-r--r--@  1 ericvn  staff   487B Apr  5 12:33 .gitattributes
```

Let me check locally again and see if I can figure more out.

ericvn:
I'm running it in istio.io, but it looks like the .* is also trying to copy the parent directory, `..` , as well? Maybe it's a Mac thing.

howardjohn:
Ah, copy `..` is definitely not good :-) seems to work for me, so yeah maybe a mac thing. I think we want .gitattributes at least, one way or another.

ericvn:
My failure is:
```
> make update-common 
cp: warning: source directory '/tmp/tmp.zktyTM2ydz/common-files/files/.' specified more than once
cp: preserving permissions for ‘/work/.git/objects/pack/pack-c570b2169c952b64c6fb434628f7fb39ab3d23a3.pack’: Permission denied
cp: preserving permissions for ‘/work/.git/objects/pack/pack-c570b2169c952b64c6fb434628f7fb39ab3d23a3.idx’: Permission denied
cp: will not create hard link '/work/files' to directory '/work/.'
make[1]: *** [common/Makefile.common.mk:103: update-common] Error 1
make: *** [update-common] Error 2
```

Good to know it might be a Mac issue. Maybe I call out .gitattributes as well as the .devcontainer. It'd be curious what happens if I run a change which should copy .gitattributes locally on a Mac and if it's ignored (while it's not in the pipeline).

ericvn:
@howardjohn So I update the first comment to discuss the issue. It seems that on a Mac if we want to copy the dot-files we need to call them out. I think we can do two things in this PR:
1: Make things work correctly for a Mac and not cause a corrupt gh type issue by changing .* to include the two dot-files in the cp.
2: Remove the .* on the cp which prevents corrupting the repo on a Mac, but a Mac won't update the two dot-files if `make update-common` is run locally.

For now, I've gone with option 1.

howardjohn (commented):
Personally the old one looked better, but if it breaks mac and we care about that (you could just run it in container? or is that not supported), this is equivilent (until we add a new file) so fine with this as well

howardjohn (approved):
add a hold, your choice

ericvn:
Thanks John, I am running it is a container inside Docker Desktop, so you would think the behavior would be the same across platforms, so it's actually something to do with Docker Desktop. Maybe something with how they figure out the bind mount.

PR #840, opened 2023-05-23
Update .golangci.yml from istio/istio

ericvn:
istio/istio/common/.golangci.yaml has changes that haven't been moved into common-files. This takes what is is istio/istio today.


ericvn:
Watching https://github.com/istio/istio/pull/45060 to see if this is needed.

howardjohn (changes requested):
This is part. of the per-repo config we added. Its intentionally only. in Istio and preserved during updates to commonfiles

ericvn:
I missed that change.

PR #844, opened 2023-06-06
upgrade depguard to v2

zirain:
related to https://github.com/istio/tools/pull/2490

### DIFF
--- a/files/common/config/.golangci.yml
+++ b/files/common/config/.golangci.yml
@@ -7,7 +7,7 @@
 
 service:
   # When updating this, also update the version stored in docker/build-tools/Dockerfile in the istio/tools repo.
-  golangci-lint-version: 1.52.x # use the fixed version to not introduce new linters unexpectedly
+  golangci-lint-version: 1.53.x # use the fixed version to not introduce new linters unexpectedly
 run:
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   deadline: 20m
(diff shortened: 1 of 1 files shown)


zirain:
/hold

wait https://github.com/istio/tools/pull/2490

files/common/config/.golangci.yml
```diff
@@ -231,8 +231,78 @@ linters-settings:
...
+          - "!**/pkg/url/**"
+          - "!pkg/test/framework/**"
+          - "!tests/fuzz/**"
+        deny:
+          - pkg: istio.io/istio/operator
+            desc: "operator should not be imported"
+          - pkg: istio.io/istio/istioctl
+            desc: "istioctl should not be imported"
+    # configuration for depguard
     packages-with-error-message:
```
zirain, line 278: delete these lines after build-image upgraded

files/common/config/.golangci.yml
```diff
@@ -231,8 +231,78 @@ linters-settings:
       # - unnamedResult
       # - wrapperFunc
   depguard:
+    # configuration for depguard v2
```
ericvn, line 234: I believe we only want the small snippet here. Although we may want to combine this change with the automated change of the build-image.

The larger snippet gets added by the override in Istio/istio.

ericvn, line 234: And are all these lines needed for each and every repo, or can you get by with the single line you had earlier? I suspect you can get by with the single line in all but the isti/istio repo which might need a bigger set (what is here). In that case, the bigger set should go in the istio/istio config.override file.
And maybe if there is a small change needed for istio.io, then we create an override there as well.

zirain: updated

files/common/config/.golangci.yml
ericvn: thanks. I think this is close to what I test in the other repos. I'll give it a another try in them in the morning and if things work, I think we can try merging this.

howardjohn:
TBH I don't really get the problem, lets just merge something :-)

ericvn:
OK. Removing DNM.

ericvn (approved):

PR #849, opened 2023-06-08
Update devcontainer.json

jacob-delgado:


jacob-delgado:
cc @therealmitchconnors we should try and get istio/tools to put up a change for this file as well in the postsubmit

howardjohn (approved):

PR #894, opened 2023-07-28
Fixing tracing in periodic jobs

howardjohn:


files/common/scripts/tracing.sh
```diff
@@ -37,8 +37,10 @@ function _genattrs() {
   # No upstream standard, so copy from https://github.com/jenkinsci/opentelemetry-plugin/blob/master/docs/job-traces.md
   if [[ -n "${PULL_NUMBER:=}" ]]
```
jwendell, line 38: Should this be `${PULL_NUMBER:-}` ?

jwendell (approved):

jwendell:
feel free to remove DNM label

files/common/scripts/tracing.sh
howardjohn: Not sure, this part seems to work fine though? I never understand all these bash tricks :grimacing:

PR #948, opened 2023-10-31
Update the license file to be unmodified apache license

syw14:
Related to https://github.com/istio/istio/pull/47651/files

The "LICENSE" file is supposed to be our inclusion of an exact copy of https://www.apache.org/licenses/LICENSE-2.0.txt as required by section 4(a) of the Apache 2 license.

### DIFF
--- a/files/LICENSE
+++ b/files/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2016-2022 Istio Authors
+   Copyright {yyyy} {name of copyright owner}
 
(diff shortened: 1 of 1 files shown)


ericvn (approved):
Thanks. Sorry for not catching the earlier change needing to be here. This will update the LICENSE file in a subset of the istio repos.

syw14:
@ericvn Hi Eric, I noticed this PR is not yet released to 1.19. I'm wondering if we could cherry-pick this and https://github.com/istio/istio/pull/47651 and release right away? thanks! (I wanted to add a cherry-pick label but seems I don't have the permission)

PR #960, opened 2023-12-12
Allow ignore findfiles

hanxiaop:
For https://github.com/istio/tools/pull/2751 to exclude releasenote template.

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -19,7 +19,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses -o -path ./vendor \) -prune -o -type f
+ifndef FINDFILES
+	FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses -o -path ./vendor \) -prune -o -type f
+endif
+
 XARGS = xargs -0 -r
 
 lint-dockerfiles:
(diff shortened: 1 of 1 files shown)


jwendell (commented):
Is it that hard to make these templates pass the linter? Is there a way to bypass the linter with some special comment like go linter has `//nolint` ?

files/common/Makefile.common.mk
```diff
@@ -19,7 +19,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses -o -path ./vendor \) -prune -o -type f
+ifndef FINDFILES
```
dhawton, line 22: Why not use ```FINDFILES ?=``` instead of the conditional?

ericvn:
I would be inclined to simply add the `./cmd/gen-release-notes/templates` to the list of paths to ignore. Some of the paths listed are repo specific. To be more elaborate, you could specify an additional variable in the Makefile.core.mk like `MARKDOWN_LINT_IGNORE=./cmd/gen-release-notes/templates` and then here add that variable into the command.

hanxiaop:
> Is it that hard to make these templates pass the linter? Is there a way to bypass the linter with some special comment like go linter has `//nolint` ?

I don’t think the markdown lint tool has similar features. Maybe there's an ignore file can be specified but it will not be repo-specific. I think @ericvn 's suggestion of FINDFILES_IGNORE makes sense.

ericvn (approved):

ericvn:
Thanks

PR #971, opened 2024-01-18
Make kubeconfig file writable

jwendell:
We are facing an error when invoking kind within the container:

```
 ...

 ✓ Installing StorageClass 💾
 • Ready after 15s 💚
ERROR: failed to create cluster: failed to write KUBECONFIG: open /config/8041fbc5: read-only file system

 ...
```


ericvn:
Do we really want the build container updating the host file? At least I believe that would be allowed to happen in this scenario. Need to investigate that some more. Maybe we want to make a copy that the container is allowed to update.

jwendell:
Right. At least it doesn't break anything: https://github.com/istio/istio/pull/48850

jwendell:
Another approach: https://github.com/istio/common-files/pull/972

PR #972, opened 2024-01-18
Make kubeconfig file writable

jwendell:
We are facing an error when invoking kind within the container:
```
 ...

  ✓ Installing StorageClass 💾
  • Ready after 15s 💚
 ERROR: failed to create cluster: failed to write KUBECONFIG: open /config/8041fbc5: read-only file system

 ...
```

Make a copy of the original file and bind it in the container, read-write, so that changes made within the container doesn't affect the original one.

### DIFF
--- a/files/common/scripts/setup_env.sh
+++ b/files/common/scripts/setup_env.sh
@@ -131,7 +131,7 @@ fi
 
 # This function checks if the file exists. If it does, it creates a randomly named host location
 # for the file, adds it to the host KUBECONFIG, and creates a mount for it. Note that we use a copy
-# of the original file, so that the container can write on it.
+# of the original file, so that the container can write to it.
 add_KUBECONFIG_if_exists () {
   if [[ -f "$1" ]]; then
     local local_config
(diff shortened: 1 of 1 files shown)


jwendell:
Dry-run: https://github.com/istio/istio/pull/48852

ericvn:
I do like this better since it prevents an update in the container from propagating back to the host's kubeconfig. I wouldn't expect that making it writeable would have an effect on what the pipeline/community does as we would have to work around/ignore the writeable problem you found (ie, this change shouldn't have an immediate negative impact on things). I'm more worried about someone in the future running some code which messes with the host (and may or may not be found and detrimental).

files/common/scripts/setup_env.sh
```diff
@@ -130,12 +130,17 @@ fi
 # echo ${CONDITIONAL_HOST_MOUNTS}
 
 # This function checks if the file exists. If it does, it creates a randomly named host location
-# for the file, adds it to the host KUBECONFIG, and creates a mount for it.
+# for the file, adds it to the host KUBECONFIG, and creates a mount for it. Note that we use a copy
+# of the original file, so that the container can write on it.
```
ericvn, line 134: nit:
```suggestion
# of the original file, so that the container can write to it.
```

ericvn (approved):

PR #991, opened 2024-02-14
Add Google's Artifact Registry to trustedRegistries

sergii-ssh:


files/common/config/.hadolint.yml
```diff
@@ -13,3 +13,4 @@ trustedRegistries:
   - gcr.io
   - docker.io
   - quay.io
+  - us-docker.pkg.dev
```
howardjohn, line 16: nit: probably no need to only trust 'us'?

howardjohn (approved):

files/common/config/.hadolint.yml
sergii-ssh: Updated to multi-region - https://github.com/istio/common-files/pull/992

PR #1011, opened 2024-04-01
kind: Respect IP_FAMILY settings for non-default config as well

unicell:
Currently IP_FAMILY environment variable is mutually exclusive with `--kind-config` and only works with default kind config. This makes it less useful to test IPv6 or dualstack with other kind config combinations.

This change makes IP_FAMILY an independent setting that works for both default and custom kind config specified.

### DIFF
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -172,14 +172,15 @@ function setup_kind_cluster() {
   if [[ -z "${CONFIG}" ]]; then
     # Kubernetes 1.15+
     CONFIG=${DEFAULT_CLUSTER_YAML}
-    # Configure the cluster IP Family only for default configs
-    if [ "${IP_FAMILY}" != "ipv4" ]; then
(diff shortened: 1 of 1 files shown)


files/common/scripts/kind_provisioner.sh
```diff
@@ -172,14 +172,15 @@ function setup_kind_cluster() {
   if [[ -z "${CONFIG}" ]]; then
     # Kubernetes 1.15+
     CONFIG=${DEFAULT_CLUSTER_YAML}
-    # Configure the cluster IP Family only for default configs
-    if [ "${IP_FAMILY}" != "ipv4" ]; then
-      grep "ipFamily: ${IP_FAMILY}" "${CONFIG}" || \
-      cat <<EOF >> "${CONFIG}"
+  fi
+
+  # Configure the cluster IP Family only for default configs
```
ericvn, line 177: This comment is now incorrect.

unicell:
@ericvn Mind taking another look and let me know if there's anything else needed? Thank you!

ericvn (approved):
Thanks

PR #1023, opened 2024-04-16
Add option for local caching

keithmattix:
If the build container ever gets stopped/updated, the local cache is lost. This PR adds a flag to persist the local cache to the user's .cache directory to save expensive artifacts across builds (especially for bazel builds)

### DIFF
--- a/files/common/scripts/run.sh
+++ b/files/common/scripts/run.sh
@@ -33,11 +33,19 @@ source "${WD}/setup_env.sh"
 
 MOUNT_SOURCE="${MOUNT_SOURCE:-${PWD}}"
 MOUNT_DEST="${MOUNT_DEST:-/work}"
+PERSIST_LOCAL_CACHE=${PERSIST_LOCAL_CACHE:-false}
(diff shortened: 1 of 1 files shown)


files/common/scripts/run.sh
```diff
@@ -53,9 +61,9 @@ read -ra DOCKER_RUN_OPTIONS <<< "${DOCKER_RUN_OPTIONS:-}"
     -e IN_BUILD_CONTAINER=1 \
     -e TZ="${TIMEZONE:-$TZ}" \
     --mount "type=bind,source=${MOUNT_SOURCE},destination=/work" \
+    ${CACHE_MOUNT} \
     --mount "type=volume,source=go,destination=/go" \
     --mount "type=volume,source=gocache,destination=/gocache" \
-    --mount "type=volume,source=cache,destination=/home/.cache" \
```
howardjohn, line 58: how does this get lost? Its a persistent volume

keithmattix: 🤦🏾 PEBKAC; my bad

PR #1025, opened 2024-04-21
golangci-lint: enable testifylint linter

mmorel-35:
golangci-lint: enable [testifylint](https://github.com/Antonboom/testifylint) linter

It will help test quality and clarity by following testify best practices


howardjohn:
What's the goal of this? We basically don't use testify

PR #1032, opened 2024-04-25
Add ocp_setup.sh to set common setup for OCP clusters

fjglira:
Adding `ocp_setup.sh` similar to what `kind_provisioner.sh` does. The idea is to have common setup in this file to be used later to run `istio/istio` and `istio-ecosystem/sail-operator` integration test over OCP clusters using the proper script on each repo

### DIFF
--- a/files/common/scripts/ocp_setup.sh
+++ b/files/common/scripts/ocp_setup.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+
+# WARNING: DO NOT EDIT, THIS FILE IS PROBABLY A COPY
+#
+# The original version of this file is located in the https://github.com/istio/common-files repo.
+# If you're looking at this file in a different repo and want to make a change, please go to the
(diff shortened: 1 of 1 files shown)


dgn:
> Can you explain a bit more? The T&R stance currently is to test `istio` organization exclusively on kind, so we would want to get more buy-in before making any alternative decisions.
> 
> In the meantime, since that leaves only `istio-ecosystem/sail-operator` using OCP it probably would be more expedient to check this into sail-operator directly?

For some context, we forked the istio repository into our org to run tests against OCP there. We basically pull in the latest commits regularly, spin up a cluster and run the tests. So we thought that maybe contributing them upstream would make sense - we would be maintaining them, otherwise our CI jobs fail. And if others are interested to locally test against OCP, they could do it straight out of the istio repo. But if you think it should be located elsewhere, we can keep it in the fork for now.

fjglira:
This is not needed anymore, we are handling this in a fork

PR #1084, opened 2024-08-20
update kind-provisioner.sh to stop mutating kind config files

ilrudie:
This is intended to stop kind_provisioner.sh from mutating kind config files on disk. I did add some defaulting of IP_FAMILY here even though integ-suite-kind.sh also defaults this. I am not sure if anyone uses the provisioner in and of itself but I suspect someone does and adding defaulting here should keep their UX the same, near as I can tell at least.

https://github.com/istio/istio/pull/52780 is the same change in istio/istio to be able to test for any impact. Please comment there to kick off any integ tests which you have concerns this change may impact.


bleggett (approved):
ty!

jacob-delgado (approved):

PR #1141, opened 2025-01-23
Ignore proto errors temporarily until the envoy dns refactor is complete

keithmattix:


howardjohn (commented):
can we not just //nolint on their usage? or if not there is a repo-specific lint config we merge into the common one this probably makesore sense in

keithmattix:
Works for me; I just opened this in case we didn't like the repo-specific stuff

mikemorris:
I think we may need to reopen this or find another way to configure it, modifying the copy directly gets overwritten by release machinery, see https://github.com/istio/istio/pull/54832/files#r1927806137

PR #1151, opened 2025-02-26
[DNM] Build container needs `SYS_ADMIN`

bleggett:
https://github.com/istio/istio/pull/55228 added this in-repo but it needs to go in common-files.

I'm not entirely sure we need to or want to run the whole build container with `SYS_ADMIN` tho - we had this same problem over in istio/ztunnel.

I'm opening this just for tracking/visibility purposes since the change already entered `master`.

### DIFF
--- a/files/common/scripts/run.sh
+++ b/files/common/scripts/run.sh
@@ -47,6 +47,7 @@ read -ra DOCKER_RUN_OPTIONS <<< "${DOCKER_RUN_OPTIONS:-}"
     "${DOCKER_RUN_OPTIONS[@]}" \
     --init \
     --sig-proxy=true \
+    --cap-add=SYS_ADMIN \
     ${DOCKER_SOCKET_MOUNT:--v /var/run/docker.sock:/var/run/docker.sock} \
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

bleggett:
We do not need this if we do https://github.com/istio/istio/pull/55276, so I'd prefer to do that instead of this.

But leaving up for comment.

bleggett:
> We do not need this if we do [istio/istio#55276](https://github.com/istio/istio/pull/55276), so I'd prefer to do that instead of this.
> 
> But leaving up for comment.

So  I was incorrect here - 

Before https://github.com/istio/istio/pull/54845 the unit tests that did `unshare` would fail, but `make gen` (which runs the tests) would work within the build container without elevated perms.

After https://github.com/istio/istio/pull/54845 the unit tests that did `unshare` would fail, but `make gen` (which runs the tests) stopped working, because the way the tests failed changed.

The unshare will fail in either case.

Given that, this is simplest since it will fix the unshare and the `make gen`

PR #1190, opened 2025-06-12
exclude node_modules from getting linted

Ajay-singh1:
This PR excludes the node_modules to be linted by the typescript linter

### DIFF
--- a/files/common/Makefile.common.mk
+++ b/files/common/Makefile.common.mk
@@ -59,7 +59,7 @@ lint-sass:
 	@${FINDFILES} -name '*.scss' -print0 | ${XARGS} sass-lint -c common/config/sass-lint.yml --verbose
 
 lint-typescript:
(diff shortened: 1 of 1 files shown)


files/common/Makefile.common.mk
```diff
@@ -59,7 +59,7 @@ lint-sass:
 	@${FINDFILES} -name '*.scss' -print0 | ${XARGS} sass-lint -c common/config/sass-lint.yml --verbose
 
 lint-typescript:
-	@${FINDFILES} -name '*.ts' -print0 | ${XARGS} tslint -c common/config/tslint.json
+	@${FINDFILES} -name '*.ts' -not -path "./node_modules/*"  -print0 | ${XARGS} tslint -c common/config/tslint.json
```
dhawton, line 62: Should this more generic to exclude all node_modules directories, not just those in the root dir? Similar to the format-go target below that excludes all gen.go and pb.go files?

PR #1196, opened 2025-06-27
Relabel bind mounts for SELinux

skitt:
When SELinux is enabled and enforcing, bind mounts need to be relabeled to work correctly in many cases. This is configured by adding the "z" option to --volume definitions (it is not supported with --mount type=bind).

This adds a relabeling function to run.sh which adds the appropriate option to --volume definitions, and warns if it finds a --mount type=bind option. It also converts --mount type=bind declarations to the corresponding --volume definition.

### DIFF
--- a/files/common/scripts/run.sh
+++ b/files/common/scripts/run.sh
@@ -39,26 +39,59 @@ read -ra DOCKER_RUN_OPTIONS <<< "${DOCKER_RUN_OPTIONS:-}"
 [[ -t 0 ]] && DOCKER_RUN_OPTIONS+=("-it")
 [[ ${UID} -ne 0 ]] && DOCKER_RUN_OPTIONS+=(-u "${UID}:${DOCKER_GID}")
 
+selinux_relabel() {
+    if ! selinuxenabled 2>/dev/null; then
+	# SELinux is not enabled, no processing
+	printf "%s " "$@"
+	return
+    fi
+
+    local arg volume
+    volume=false
+    for arg; do
+	if $volume; then
+	    if [[ "$arg" =~ .*:.*:.* ]]; then
+		printf "%s,z " "$arg"
+	    else
+		printf "%s:z " "$arg"
+	    fi
+	    volume=false
+	elif [ "$arg" = --volume ] || [ "$arg" = "-v" ]; then
+	    printf "%s " "$arg"
+	    # Process the next argument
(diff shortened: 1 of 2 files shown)


files/common/scripts/run.sh
```diff
@@ -39,26 +39,59 @@ read -ra DOCKER_RUN_OPTIONS <<< "${DOCKER_RUN_OPTIONS:-}"
...
+
+    local arg volume
+    volume=false
+    for arg; do
+	if $volume; then
+	    if [[ "$arg" =~ .*:.*:.* ]]; then
+		printf "%s,z " "$arg"
+	    else
+		printf "%s:z " "$arg"
+	    fi
```
fjglira, line 57: Using `z` sets shared relabel, so this means that the host path SELinux label is changed and any container can access it, right? Because this is only used to start the build container, wouldn't it be better to use `Z` for private relabeling — so the label is exclusive to this container?

fjglira (commented):
It looks good to me, just a minor question. If anyone is against this I'll approve the PR

files/common/scripts/run.sh
skitt: With `Z`, it’s impossible to run multiple containers in parallel if they need access to the same host directories. Would that be a problem in practice? Probably not; if this is *only* used for the build container, then no indeed.

The other potential issue with `Z` is that this is also used to mount the Docker socket; I haven’t been able to verify that but I suspect this would prevent other containers from using DIND.

I think the real question here is whether we want to guarantee exclusive access to the mounted volumes, in particular to avoid race conditions. For build containers that does seem desirable, I’ll update the PR.

skitt: Ah I remember why I did it this way: using `z` reproduces the existing behaviour, where mounts aren’t exclusive.

skitt: @zmiklank what do you think? I see #1293 also uses `z` rather than `Z`.

zmiklank: yes, I did use `z` on purpose. In my opinion it is more convenient for the users (e.g., running two containers mounting the same files at the same time) and I do not really see any security concerns with it.

fjglira: Ok, thanks for the clarification

fjglira (approved):
lgtm

PR #1215, opened 2025-08-28
update kind to use 1.34 as the default k8s version

jacob-delgado:
Depends on https://github.com/istio/tools/pull/3269

Will wait until test-infra updates and the build from tools has been propagated to other repos


keithmattix:
@jacob-delgado do you need me to push the new image?

keithmattix (approved):

PR #1266, opened 2026-03-20
Do not specify image arch explicitly

FilipB:


fjglira (approved):
lgtm, make sense avoid setting arm arch by default

PR #1278, opened 2026-05-06
Increase inotify limits for kind-provisioner

bmangoen:
Increasing inotify limits (`fs.inotify.max_user_instances` and `fs.inotify.max_user_watches`) to prevent `too many open files` errors with ZTunnel and ambient mode


bmangoen:
cc. @fjglira

howardjohn (commented):
isn't watching the configmap 1 fsnotify instance? Are we leaking something?

bmangoen:
> isn't watching the configmap 1 fsnotify instance? Are we leaking something?

Actually, I checked deeply and it was a local env issue. Closing this PR. @howardjohn thank you for answer

PR #1280, opened 2026-05-28
update license lint

Stevenjin8:
See https://github.com/istio/test-infra/pull/5924#issuecomment-4564022111

### DIFF
--- a/files/common/config/license-lint.yml
+++ b/files/common/config/license-lint.yml
@@ -64,6 +64,11 @@ allowlisted_modules:
 # MIT: https://github.com/ghodss/yaml/blob/master/LICENSE
 - github.com/ghodss/yaml
 
+# Apache 2.0: this is a sibling module within this repo (see top-level go.mod
+# `replace istio.io/test-infra/tools/prowgen => ./tools/prowgen`). The license
+# linter cannot detect a license for unpublished local modules.
+- istio.io/test-infra/tools/prowgen
+
 # BSD: https://github.com/gogo/protobuf/blob/master/LICENSE
 - github.com/gogo/protobuf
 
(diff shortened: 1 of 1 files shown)


files/common/config/license-lint.yml
```diff
@@ -64,6 +64,11 @@ allowlisted_modules:
 # MIT: https://github.com/ghodss/yaml/blob/master/LICENSE
 - github.com/ghodss/yaml
 
+# Apache 2.0: this is a sibling module within this repo (see top-level go.mod
+# `replace istio.io/test-infra/tools/prowgen => ./tools/prowgen`). The license
```
jwendell, line 68: This comment is only valid for the tools repo. Remove it?

dhawton: It's valid for the test-infra repo, not tools. Not sure why you want it removed? It's needed. Unless you're referring strictly to the replace comment, in which  ya, it could be clarified it's for test-infra... see https://github.com/istio/test-infra/pull/5932 for why this is needed

jwendell: "this is a sibling module within this repo" => keep in mind this file will be propagated to many repos. I'm fine with updating the comment to make it generic if you prefer. Removing it entirely is an option too.

dhawton: We cannot remove it because it's blocking test-infra. So long as this doesn't impact other repos, I think it's fine.

jwendell: I'm talking about the comment "this is a sibling module within this repo". If I read this while browsing istio/istio for instance, I'll be confused.

Stevenjin8: @jwendell will update

jwendell (approved):

PR #1293, opened 2026-07-30
Add CONTAINER_SELINUX option for build containers on SELinux-enforcing systems

zmiklank:
When CONTAINER_SELINUX=1 is set, all bind mounts get the :z shared SELinux label, allowing containerized builds and tests to work on SELinux-enforcing systems (Fedora, RHEL, CentOS) without disabling SELinux enforcement.

Bind mounts are switched from --mount to -v syntax for universal :z support across both Docker and Podman.


zmiklank:
Hi @skitt I noticed [your PR](https://github.com/istio/common-files/pull/1196) with similar goal only after I finished the work on mine, so I decided to publish it here anyway and I can close it if preferred.

Changes in this PR are simpler and my approach requires additional code support in each repository that uses `common-files` scripts so that would be a disadvantage.

In any way it would be cool if we get selinux support merged, let me know how I can help (I do not have review permissions here though).

zmiklank:
Closing in favor of https://github.com/istio/common-files/pull/1196.

PR #1300, opened 2026-08-04
build, selinux: convert mount, use --tmpfs where needed

zmiklank:


zmiklank:
Hi @skitt, I did test the selinux support on istio repo, and it seems it would be useful to also support mount conversion and relabel, wdyt?

fjglira:
Hey @zmiklank can we check again if this is still needed

zmiklank:
Hi @fjglira I think this is indeed still needed.

PR #1302, opened 2026-08-20
Upgrade default KIND node to v1.36

jewertow:
The [support status](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases) table shows that we support k8s 1.36, but 1.35 is the highest version used in tests for now.


jewertow:
cc: @dhawton @kfaseela

kfaseela (approved):

PR #1313, opened 2026-09-04
metallb configurable hub

Stevenjin8:
This allows us to use a Pull through cache and save on networking ingress

### DIFF
--- a/files/common/scripts/kind_provisioner.sh
+++ b/files/common/scripts/kind_provisioner.sh
@@ -407,9 +407,65 @@ function install_calico {
   kubectl --kubeconfig="$KUBECONFIG" wait --for condition=ready -n kube-system pod -l k8s-app=calico-kube-controllers --timeout 90s
 }
 
+function configure_ecr_image_pull_secret() {
+  local KUBECONFIG="${1}"
+  local ecr_password
+  local xtrace_enabled=false
+
+  if [[ -z "${ECR_REGISTRY:-}" ]]; then
+    return
+  fi
+
+  if [[ "$-" == *x* ]]; then
+    xtrace_enabled=true
+    set +x
+  fi
+
+  ecr_password=$(aws ecr get-login-password)
+  kubectl --kubeconfig="$KUBECONFIG" --namespace metallb-system create secret docker-registry ecr-pull-secret \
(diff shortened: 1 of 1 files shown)


files/common/scripts/kind_provisioner.sh
```diff
@@ -407,9 +407,65 @@ function install_calico {
...
+    --dry-run=client \
+    --output=yaml | kubectl --kubeconfig="$KUBECONFIG" --namespace metallb-system apply -f -
+  unset ecr_password
+
+  if [[ "${xtrace_enabled}" == true ]]; then
+    set -x
+  fi
+}
+
+function configure_metallb_image_pull_secret() {
```
Stevenjin8, line 438: we need to create a secret because even though the prow pod has access to the pull through cache, the inner kind nodes have a different container runtime and do not have such access

files/common/scripts/kind_provisioner.sh
```diff
@@ -407,9 +407,65 @@ function install_calico {
...
+  if [[ -z "${ECR_REGISTRY:-}" ]]; then
+    return
+  fi
+
+  if [[ "$-" == *x* ]]; then
+    xtrace_enabled=true
+    set +x
+  fi
+
+  ecr_password=$(aws ecr get-login-password)
```
Stevenjin8, line 424: this is an RO token that is valid for 12 hours. the pull through cache is not accessible from outside the cluster: https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login-password.html

files/common/scripts/kind_provisioner.sh
```diff
@@ -407,9 +407,65 @@ function install_calico {
...
+- name: ecr-pull-secret
+EOF
+  done
+
+}
+
 function install_metallb() {
   KUBECONFIG="${1}"
-  kubectl --kubeconfig="$KUBECONFIG" apply -f "${COMMON_SCRIPTS}/metallb-native.yaml"
+  if [[ -n "${ECR_REGISTRY:-}" ]]; then
```
Stevenjin8, line 461: ECR_REGISTRY is our pull through cache

keithmattix (approved):

